PR #1, opened 2016-10-19
Add Documentation to repo

RichardLitt:
This adds a License foe this repo, a contributing document for the entire organization, and a README filled out with content. 

I would like to add captains to each repo, too, if that is good. I can write up something about captainship as we are using it in multiformats, too. For that, I expect that I would actually be the captain for this repo, as nicola should focus on the specs and not on doing random PR merges.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,53 @@
-# ipld
-InterPlanetary Linked Data
+# IPLD
+
+[![](https://img.shields.io/badge/made%20by-Protocol%20Labs-blue.svg?style=flat-square)](http://ipn.io)
+[![](https://img.shields.io/badge/project-ipld-blue.svg?style=flat-square)](http://github.com/ipld/ipld)
+[![](https://img.shields.io/badge/freenode-%23ipfs-blue.svg?style=flat-square)](http://webchat.freenode.net/?channels=%23ipfs)
+
+> InterPlanetary Linked Data
+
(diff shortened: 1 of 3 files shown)


RichardLitt:
cc @diasdavid

README.md
```diff
@@ -1,2 +1,53 @@
...
-InterPlanetary Linked Data
+# IPLD
+
+[![](https://img.shields.io/badge/made%20by-Protocol%20Labs-blue.svg?style=flat-square)](http://ipn.io)
+[![](https://img.shields.io/badge/project-ipld-blue.svg?style=flat-square)](http://github.com/ipld/ipld)
+[![](https://img.shields.io/badge/freenode-%23ipfs-blue.svg?style=flat-square)](http://webchat.freenode.net/?channels=%23ipfs)
+
+> InterPlanetary Linked Data
+
+Note that, as in many IPFS repositories, most of the work is happening in [the issues](https://github.com/ipld/specs/issues/) or in [active pull requests](https://github.com/ipld/specs/pulls/). Go take a look!
```
VictorBjelkholm: This is linking to `ipld/specs` instead of `ipld/ipld`

diasdavid:
Why isn't the License MIT (I might be totally unaware of where that decision was made).

> I would like to add captains to each repo, too, if that is good. I can write up something about captainship as we are using it in multiformats, too

I would like that too, but preferably in a separate PR

RichardLitt:
Because MIT only refers to software. This is the corollary open license for docs. 

Will open a separate PR for that.

Issue #33, opened 2018-05-12
Update the README with the newly created tutorials, demos and tools


diasdavid:
@alanshaw and @olizilla you have shipped multiple demos of IPLD in the last few months. Do you feel like sharing on the README to be entry points for others?

PR #35, opened 2018-05-14
docs: add CID inspector link

alanshaw:
refs #33


vmx:
BTW: The JS commit message guidelines don't really apply here as this is not JS and it will always be "docs" i guess. So feel free to leave that `docs:` out. But having it is also totally fine.

PR #37, opened 2018-05-29
docs: invite folks to the IRC channel

diasdavid:
Perhaps it should another color :)


Stebalien:
go-ipfs uses blue but using different colors is probably a good idea.

But in the interest of not leaving PRs open forever, I'll merge this and we can change it later if we decide on some consistent color scheme.

PR #72, opened 2019-05-28
events: add 2019 IPLD summit @ Berlin notes

rvagg:
cc: @mikeal @vmx @warpfork @rvagg @aschmahmann @Raynos

I didn't catch names of other people that attended day 4, if others did could you please edit this and add them to the doc?

I've pulled together what I had, some of the notes by @warpfork that he captured (unfortunately we lost the excellent proposed ipld/schemas directory layout that was mapped out).

Please feel free to edit this branch if you have access or add your thoughts inline about some of the highlights of the discussion and add any other artifacts that you may have.

### DIFF
--- a/events/2019-ipld-summit-berlin-notes.md
+++ b/events/2019-ipld-summit-berlin-notes.md
@@ -1,0 +1,200 @@
+# 2019 IPLD Summit Notes
+
+May 2019, Berlin
+
+IPLD Team:
+ - @mikeal
+ - @vmx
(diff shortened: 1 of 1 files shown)


warpfork:
This is ancient and I think we collectively forgot to push it to merge!  But it looks like fine and truthful notes to me.  Let's merge it?

I'll do so in about a day if no one objects.

vmx (approved):

PR #84, opened 2020-02-11
Bring entry point docs inline with reality

mikeal:
We’re very overdue for an update here.

I’ve made some basic changes and started a bit of a new outline, would like to get input and contributions from the team before moving forward.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,14 +6,13 @@
 
 [![](https://img.shields.io/badge/made%20by-Protocol%20Labs-blue.svg?style=flat-square)](https://protocol.ai)
 [![](https://img.shields.io/badge/project-ipld-orange.svg?style=flat-square)](https://github.com/ipld/ipld)
-[![](https://img.shields.io/badge/freenode-%23ipld-orange.svg?style=flat-square)](https://webchat.freenode.net/?channels=%23ipld)
 
 ### Welcome to the Internet of Data-Structures
 
 IPLD is a set of standards and implementations for creating decentralized data-structures that are universally addressable and linkable. These structures will allow us to do for data what URLs and links did for HTML web pages.
 
-- [Enter IPLD](#enter-ipld)
 - [Docs](#docs)
+- [Stacks](#stacks)
 - [Examples and Tutorials](#examples-and-tutorials)
 - [Tools](#tools)
 - [Implementations](#implementations)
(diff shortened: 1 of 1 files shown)


rvagg:
can + worms

How committed are you to this right now? I think the best thing you could do is invest in some quality introductory text, something we can reuse at all our main entry points.

That "Codec Implementations" could be dropped, doing it across language isn't useful when you think of the audience (i.e. they're mostly going to be language-specific). So the per-language sections are a good idea but they're going to need to have a short explanation of the flux that's going on in each of them. "Existing JS stuff start here ... have a look at newer stuff here ..." which could be very similar for Go. But you have to explain why we're not evolving older stuff and replacing it all with newer things and give a good explanation of the state of that work .. and then we need to keep it updated! But I suppose we need to do that somewhere anyway.

warpfork:
quick thoughts:

- +1 to @rvagg 's comments
- i did a quick zeerust check on he "enter the merkle forest" talk and... it's not too bad (though that's maybe surprising since we've certainly done a lot of work since then!).  We can see key terms like "Data Model" shining through: https://youtu.be/Bqs_LzBjQyk?t=858 .  So while I'd love to have a brief, modernized overview video, this one's still not too bad.  (Also room for improvement: for example, this video spends a lot of time talking about multiformats, which is probably not where I'd spend most of the airtime if we did a new IPLD-focused explainer.)
- I like the idea of segmenting up the "implementations" section a bit more, and if we're going to list codecs, then definitely giving codecs their own berth because there's so very many of them.  (Follow-on thought: For ecosystemic legibility, it might be pretty great if we had a clear one-stop-shop central library per language -- this is basically what I'm aiming for in go-ipld-prime.  Broken out into sub-packages, sure, but with a clear unity and clear single starting point that has all the key interfaces at once.  Does this jive in the JS POV?)
- Definitely true that we need some plan of action for getting the "resolver" links outta there.  That should... probably move to a "legacy" heading.  With some prejudice.  It would be good to sharply reduce the number of people who come into our docs and encounter those old interfaces.

README.md
```diff
@@ -22,17 +21,20 @@ IPLD is a set of standards and implementations for creating decentralized data-s
...
 
-**Watch the Merkle Forest Talk**
+- [IPLD Specs](https://github.com/ipld/specs)
+- [CID spec](https://github.com/multiformats/cid)
 
-[![](/img/enter-merkle-forest.jpg)](https://www.youtube.com/watch?v=Bqs_LzBjQyk)
+## Stacks
 
-## Docs
+IPLD has a [layer model]() and is much more than "one thing." As such, implementations of IPLD aren't single libraries but stacks of libraries that vary a bit by programming language.
```
vmx: There's the actual link target missing.

mikeal: yup, want to work out the structure before filling in all the details.

mikeal:
> I also want the link to IRC back.

I want to get the full discourse info in there instead of just IRC.

> video

The video isn’t *that* out of date, but I’d just prefer not to have a video as the first thing you see. We should strive for some introductory text that can be quickly read that gives the reader all the important points they would have gotten from a 20 minute talk. The talk should still be referenced later in the resources section though.

> codecs

Good points on the codec implementation list.

Maybe a better approach would be to link from here to each language’s entry point (new and old entry points when relevant) and *those* README’s can list the codec implementations if they think that’s valuable.

warpfork:
> video

> I’d just prefer not to have a video as the first thing you see. We should strive for some introductory text that can be quickly read that gives the reader all the important points they would have gotten from a 20 minute talk. 

yeah, Okay, that I definitely agree with.  I'm definitely of the type of person that gets supremely irritated when the only explanation of something I can find is video format.

PR #94, opened 2021-02-16
document the proper use of code coverage tools

marten-seemann:
While code coverage tools are a valuable tool during development, having code coverage aggressively enforced in CI is not a good idea: Removing a bunch of well-tested, but ultimately unnecessary code from a project would be regarded as an improvement by all developers, however, a code coverage tool would only see a drop in coverage and fail the status check on the PR.

We can use "non-blocking status checks" config from the Codecov manual: https://docs.codecov.io/docs/common-recipe-list#set-non-blocking-status-checks.

What we want the code coverage tool to do is report the changes, but to not affect the PR status. See here for a real-world example how this will look like: https://github.com/marten-seemann/codecov-test/pull/5

### DIFF
--- a/best-practices/go-ci.md
+++ b/best-practices/go-ci.md
@@ -27,6 +27,23 @@ should also run on 1.15.x. Note that upstream Go supports
 [two major releases](https://github.com/golang/go/wiki/Go-Release-Cycle#release-maintenance)
 at once as well.
 
+### Using Code Coverage Tools
+
+Code coverage can be determined in CI, but should not be used as an automatic status checks.
+
+When using Codecov, this can be achieved by using the following `codecov.yml`:
+
+```yml
+coverage:
+  status:
+    project:
+      default:
+        informational: true
+    patch:
+      default:
+        informational: true
+```
+
 #### How to use this document at Protocol Labs
 
 Marten Seemann and the IPLD team have set up [a template repository](https://github.com/ipld/.github)


best-practices/go-ci.md
```diff
@@ -27,6 +27,23 @@ should also run on 1.15.x. Note that upstream Go supports
 [two major releases](https://github.com/golang/go/wiki/Go-Release-Cycle#release-maintenance)
 at once as well.
 
+### Using Code Coverage Tools
+
+Code coverage should be determined in CI, but should not be used as an automatic status checks.
```
mvdan, line 32: `s/should be/can be/`, maybe? I don't think we necessarily want to encourage people to use CI for it. one could collect the same data locally.

best-practices/go-ci.md
```diff
@@ -27,6 +27,23 @@ should also run on 1.15.x. Note that upstream Go supports
...
+
+When using Codecove, this can be achieved by using the following `codecov.yml`:
```
mvdan, line 34: s/Codecove/Codecov/

mvdan (approved):

best-practices/go-ci.md
marten-seemann: Makes sense.

best-practices/go-ci.md
marten-seemann: Fixed.

PR #101, opened 2021-06-18
Promote CARv1 spec, add a test fixture

rvagg:
1. this is against `rvagg/package.json`, I think #97 should be merged to do npm properly
2. I don't know how to enable download of the `.car` file in the test fixture I've included here

I added a test fixture generator as an example in js-car @ https://github.com/ipld/js-car/pull/30, which will also generate the same format output for a CARv2 which I'll add with the draft spec as the next PR here.


rvagg:
rebased against `master`, good to go now, nice and clean

rvagg:
including the carv2 work from #107

PR #103, opened 2021-06-21
data-structures: update filecoin's HAMT version

rvagg:
This took a bit of work, but it's got the 3 commits from https://github.com/ipld/specs/pull/359, which have mostly been approved. I wouldn't mind a quick eye over the schema at the bottom which the last commit merged from two separate schemas. Is this readable enough or should I back that commit out?

I'll add a tombstone to specs for this file once this is merged.

### DIFF
--- a/specs/advanced-data-layouts/hamt/spec.md
+++ b/specs/advanced-data-layouts/hamt/spec.md
@@ -29,7 +29,10 @@
   * [Security](#security)
 * [Appendix: Filecoin HAMT Variant](#appendix-filecoin-hamt-variant)
   * [Implicit and fixed parameters](#implicit-and-fixed-parameters)
+  * [Map / Bitfield layout](#map--bitfield-layout)
   * [Block layout](#block-layout)
+    * [Actors v0.9 and v2 form](#actors-v09-and-v2-form)
+    * [Actors v3+ form](#actors-v3-form)
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


mvdan (approved):
SGTM, just like the last iteration on your specs PR.

specs/advanced-data-layouts/hamt/spec.md
```diff
@@ -297,27 +300,52 @@ The Filecoin HAMT _does not_ use an explicit root block (`HashMapRoot`) to encod
...
 
+### Map / Bitfield layout
+
+Rather than encoding the HAMT node's `map` field as the full sequence of byte-to-element mapping with little-endian byte addressing, the Filecoin HAMT uses the byte layout of a Go [`big.Int`](https://golang.org/pkg/math/big/#Int) as a bitfield to represent the HAMT node's `map`. In practice this means that the byte array length can vary depending on which bits are set within the bitfield. Implementations must use the same encoding and decoding rules as `big.Int` to handle this field. See [this discussion thread](https://github.com/filecoin-project/go-hamt-ipld/issues/54#issuecomment-670314664) for an exploration of the specifics.
+
 ### Block layout
 
-An IPLD schema representing the Filecoin HAMT varies from the IPLD HashMap [schema](#schema) so any implementation needing to read Filecoin HAMT blocks will need to handle its specific layout:
+The IPLD schema describing the Filecoin HAMT varies from the IPLD HashMap [schema](#Schema) and also between version of the Filecoin Actors, so any implementation needing to read Filecoin HAMT blocks will need to handle its specific layout (both layouts if historical data is important to create / read).
```
warpfork, line 309:
```suggestion
The IPLD schema describing the Filecoin HAMT varies from the IPLD HashMap [schema](#Schema) and also between versions of the Filecoin Actors, so any implementation needing to read Filecoin HAMT blocks will need to handle its specific layout (both layouts if historical data is important to create / read).
```

Just a missing 's'.

specs/advanced-data-layouts/hamt/spec.md
```diff
@@ -297,27 +300,52 @@ The Filecoin HAMT _does not_ use an explicit root block (`HashMapRoot`) to encod
...
+
+The **v3** form of the Filecoin HAMT has the same IPLD layout as the IPLD HashMap except for the lack of an explicit root node to describe the parameters. A Filecoin HAMT's root is the equivalent of `HashMapNode`, which is the same as all other nodes and parameters are implicit.
```
warpfork, line 311:
```suggestion
The **v3** form of the Filecoin HAMT has the same [substrate](/glossary/#substrate) Data Model layout as the IPLD HashMap except for the lack of an explicit root node to describe the parameters. A Filecoin HAMT's root is the equivalent of `HashMapNode` -- the root node is the same as all other nodes, and rather than parameters being explicit, the parameters are implicit and must be derived from the ambient Filecoin chain version.
```

Proposal of wording to make it say "IPLD" slightly fewer times in a row, and use the 'substrate' term instead (which also has a glossary entry, now, so we can link to that).  Also proposed a little more expansion about the parameter implication thing.

warpfork (approved):
Reads pretty great to me.  Proposed a few tiny tweaks.

Confirmed Like for the combined schema at the end.  My eyes flipped back and forth a couple times, checking if the different types are really as similar as they look, but that was pretty correct: the info was all sufficiently proximate that I could do that check quickly.  Good experience.

PR #104, opened 2021-06-21
feat: allow inline link definitions for kinded unions

rvagg:
This is https://github.com/ipld/specs/pull/371, but here.

---

e.g.

```
type Foo union {
  | &Blip link
  | Blop string
} representation kinded
```

rather than having to `type BlipLink &Blip` before using it in a kinded union.

---

I've been using this pattern already, e.g. https://github.com/ipld/specs/blob/9bab36f7c256ae8ff14e0190ff0327e631ef2e15/data-structures/hashmap.md#L113-L116

I think I assumed we had this locked in but I guess we never got it into schema-schema. Without this, to use a link to another type you've defined, you have to make a new link type for it `type HashMapNodeLink &HashMapNode`.

### DIFF
--- a/specs/schemas/schema-schema.ipldsch
+++ b/specs/schemas/schema-schema.ipldsch
@@ -378,7 +378,15 @@ type UnionRepresentation union {
 ##
 ## The referenced type must of course produce the RepresentationKind it's
 ## matched with!
-type UnionRepresentation_Kinded {RepresentationKind:TypeName}
+type UnionRepresentation_Kinded {RepresentationKind:KindedType}
+
+## "KindedType" is used to allow inline link definitions within a kinded union
+## as a shorthand rather than requiring all links be defined as a separate type
+## before used within a kinded union.
+type KindedType union {
+  | TypeName "string"
+  | TypeLink "link"
+} representation kinded
 
 ## "Keyed" union representations will encode as a map, where the map has
 ## exactly one entry, the key string of which will be used to look up the name
--- a/specs/schemas/schema-schema.ipldsch
+++ b/specs/schemas/schema-schema.ipldsch
@@ -385,7 +385,7 @@ type UnionRepresentation_Kinded {RepresentationKind:TypeNameOrLink}
 ## before used within a kinded union.
 type TypeNameOrLink union {
   | TypeName string
-  | TypeLink link
+  | TypeLink map
 } representation kinded
 
 ## "Keyed" union representations will encode as a map, where the map has
--- a/specs/schemas/schema-schema.ipldsch
+++ b/specs/schemas/schema-schema.ipldsch
@@ -526,7 +526,9 @@ type TypeTerm union {
 
 ## InlineDefn represents a declaration of an anonymous type of one of the simple
 ## recursive kinds (e.g. map or list) which is found "inline" in another type's
-## definition.  It's the more complex option of the TypeTerm union.
+## definition.  It is also used to support expressive link declaration in
+## certain places so they don't have to be declared as their own types.
+## InlineDefn is the more complex option of the TypeTerm union.
 ##
 ## Note that the representation of this union -- `representation inline "kind"`
 ## -- as well as the keywords for its members -- align exactly with those
@@ -539,6 +541,7 @@ type TypeTerm union {
 type InlineDefn union {
 	| TypeMap "map"
 	| TypeList "list"
+	| TypeLink "link"
 } representation inline {
 	discriminantKey "kind"
 }
--- a/specs/schemas/schema-schema.ipldsch
+++ b/specs/schemas/schema-schema.ipldsch
@@ -378,14 +378,14 @@ type UnionRepresentation union {
 ##
 ## The referenced type must of course produce the RepresentationKind it's
 ## matched with!
-type UnionRepresentation_Kinded {RepresentationKind:KindedType}
+type UnionRepresentation_Kinded {RepresentationKind:TypeNameOrLink}
 
-## "KindedType" is used to allow inline link definitions within a kinded union
+## "TypeNameOrLink" is used to allow inline link definitions within a kinded union
 ## as a shorthand rather than requiring all links be defined as a separate type
 ## before used within a kinded union.
-type KindedType union {
-  | TypeName "string"
-  | TypeLink "link"
+type TypeNameOrLink union {
+  | TypeName string
+  | TypeLink link
 } representation kinded
 
 ## "Keyed" union representations will encode as a map, where the map has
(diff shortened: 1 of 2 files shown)


rvagg:
@warpfork I've added `TypeLink` to `InlineType` because without it we can't even to `type Foo [&Any]`, `type Bar {String:&Any}` or even `type Baz struct { foo &Any }` because that's an inline definition of a link type rather than using `Link`. Of course, being able to `&SomethingElse` is important in all these places too.

PTAL and sanity check.

rvagg:
going to merge this under an assumption it's correct, there's a chance we find weirdness (as with other areas, as I keep finding with my limited use of it) when we properly put schema-schema into go-ipld-prime and wire it al up /cc @mvdan

rvagg:
actually no, I can't quite bring myself to pull the trigger, I need @warpfork's eyes on this first

specs/schemas/schema-schema.ipldsch
```diff
@@ -378,7 +378,15 @@ type UnionRepresentation union {
 ##
 ## The referenced type must of course produce the RepresentationKind it's
 ## matched with!
-type UnionRepresentation_Kinded {RepresentationKind:TypeName}
+type UnionRepresentation_Kinded {RepresentationKind:KindedType}
+
+## "KindedType" is used to allow inline link definitions within a kinded union
+## as a shorthand rather than requiring all links be defined as a separate type
+## before used within a kinded union.
+type KindedType union {
```
rvagg, line 386: `
```suggestion
type TypeNameOrLink union {
```

specs/schemas/schema-schema.ipldsch
```diff
@@ -378,7 +378,15 @@ type UnionRepresentation union {
 ##
 ## The referenced type must of course produce the RepresentationKind it's
 ## matched with!
-type UnionRepresentation_Kinded {RepresentationKind:TypeName}
+type UnionRepresentation_Kinded {RepresentationKind:KindedType}
```
rvagg, line 381:
```suggestion
type UnionRepresentation_Kinded {RepresentationKind:TypeNameOrLink}
```

specs/schemas/schema-schema.ipldsch
```diff
@@ -378,7 +378,15 @@ type UnionRepresentation union {
...
+
+## "KindedType" is used to allow inline link definitions within a kinded union
```
rvagg, line 383:
```suggestion
## "TypeNameOrLink" is used to allow inline link definitions within a kinded union
```

warpfork:
This is... still not quite an accurate description though, I think?  Developing the schema-schema is so heady :weary: 

I think what we want here is more like `TypeTerm` (which, to future-proof this post, I might shortly rename to `TypeNameOrInlineDefn`) -- the kind for the second member of the union isn't link, it's map.  Why?:

- We want a kinded union so we can do the bare string in the common case it's just a `TypeName`...
- but the other type isn't actually a link kind; it's a description of the link target type!  So as it's written right now, it's not quite right :/
  - We only really need a string's worth of data, because it's just a TypeName again.  But we can't do two strings in a kinded union, so...
- We probably want to make another type that's got a map representation for the other member of this `TypeNameOrLink` type, and then wrap the link target `TypeName` in there.
  - Which is pretty analogous to how `TypeTerm` contains `InlineDefn`.

I wonder if we should just throw open the floodgates and say unions can actually use `TypeTerm` in their values.  That would mean supporting anonymous maps and anonymous lists too.  I think the reason I didn't do this at first is that at some point I considered anon maps and anon lists in unions undesirable because I was afraid it would get syntactically out of hand... but thinking it through again today, I may have been unduely worried about that; it seems like something we could do.

warpfork:
Yeah, I'm pretty sure `TypeNameOrLink` should be a union with kinds string or map.

We have a couple options though:
- option#1: The second member of the union is just `TypeLink map`.
  - Pro: solves it.
  - Pro: only one new type introduced (`TypeNameOrLink`)
  - Con: if we ever decide to evolve this area, we're pretty stuck.
- option#2:  The second member of the union is `InlineDefn`.
  - Pro: solves it.
  - Pro: only one new type introduced (`TypeNameOrLink`)
  - Mixed: means unions suddenly allow anonymous maps and anonymous lists, too.
- option#3:  The second member of the union is `InlineDefnForUnions`.  The union is keyed (like `InlineDefn`) and has only one member: `TypeLink "link"`
  - Pro: solves it.
  - Pro: super explicit.  (Doesn't expand things the way option#2 does.)
  - Pro: if we ever do decide to evolve this (including in was like just switching to `InlineDefn`, e.g. option#2), there's room.
  - Con:  two new types introduced (`TypeNameOrLink` and also `InlineDefnForUnions`).
  
Okay so: option#1 feels a little dangerous, because it hems in future evolvability.  I could either way on 2 vs 3 at the moment.

rvagg:
Good thing I got you to look at this before merging! The `link` thing is a bit embarassing but thus is the challenge of mentally stepping down a level to schema-schema.

I'm not convinced that "if we ever decide to evolve this area, we're pretty stuck." is actually true. I don't see why, if we stick with this one, the door doesn't remain open to option 2 later on, it just means we introduce that extra type and match it on `map`, expanding functionality and not blocking existing functionality. I suppose it's a problem for strictness of DMT form if we're encoding that and matching against schema-schema.. which is a problem for a lot of this.

So I don't mind option 2 at all but I think it might depend on what your go-ipld-prime codegen does with these things. Do they just become (roughly) anonymous types that you give munged names (using lots of `--` or `__`)? I suppose you're doing that already in some places so maybe this is easy? If you're fine with doing it there then I don't see a reason not to do it here as well, although it might not be something we want to encourage too much.

warpfork:
Yeah, what I meant by "if we ever decide to evolve this area, we're pretty stuck" on option#1 there is: the serial form would have the link type info immediately there, and not wrapped in the type union, which would mean if we changed it support some of the members of the type union in the future, it would be incompatable with existing serial data.  Option#2 or option#3 would solve this by introducing a union right away, so that union could be the extension point in the future.

warpfork:
After letting this stew for a while, I still feel pretty strongly that we should solve this with either option#3 or option#2.  The diff currently contains option#1 and I don't think I'm comfortable with that approach.

warpfork:
... I've started working on a new branch that's fixing a bunch of miscellaneous issues with the schema-schema that came to my attention today.  I'm going to just cherry-pick this onto that branch and massage it to a final form myself :)  I'm so glad you discovered this and got us drafting on the fix, though!

warpfork:
Think I got it.  The fix for this stuff is included over in #136 now.  Closing this one.

PR #105, opened 2021-06-21
fix: byteprefix union representation AST form fix

rvagg:
This is https://github.com/ipld/specs/pull/370, but here

---

discovered while trying to use via TypeScript @ https://github.com/ipld/js-ipld-schema/pull/37 -> https://github.com/rvagg/js-ipld-schema-validator/pull/4

### DIFF
--- a/specs/schemas/schema-schema.ipldsch
+++ b/specs/schemas/schema-schema.ipldsch
@@ -431,9 +431,7 @@ type UnionRepresentation_Inline struct {
 ## byteprefix is an invalid representation for any union that contains a type
 ## that does not have a bytes representation.
 ##
-type UnionRepresentation_BytePrefix struct {
-	discriminantTable {TypeName:Int}
-}
+type UnionRepresentation_BytePrefix {TypeName:Int}
 
 ## TypeStruct describes a type which has a group of fields of varying Type.
 ## Each field has a name, which is used to access its value, similarly to
--- a/specs/schemas/schema-schema.ipldsch
+++ b/specs/schemas/schema-schema.ipldsch
@@ -349,11 +349,11 @@ type TypeUnion struct {
 ## can be mapped onto a serialized format for the IPLD Data Model.
 ##
 ## There are five strategies that can be used to encode a union:
-## "keyed", "envelope", "inline", "byteprefix", and "kinded".
+## "keyed", "envelope", "inline", "bytesprefix", and "kinded".
 ## The "keyed", "envelope", and "inline" strategies are all ways to produce
 ## representations in a map format, using map keys as type discriminators
 ## (some literature may describe this as a "tagged" style of union).
-## The "byteprefix" strategy, only available only for unions in which all
+## The "bytesprefix" strategy, only available only for unions in which all
 ## member types themselves represent as bytes in the data model, uses another
 ## byte as the type discrimination hint (and like the map-oriented strategies,
 ## may also be seen as a form of "tagged" style unions).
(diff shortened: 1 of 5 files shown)


warpfork:
I had an alternative idea in this area we might want to consider (just as a possibility, not confidentally attached to this yet):

What if we turned this into a map `{String:TypeName}`, and used hex here?

- Puts the key in the key; mildly nice.
- Lets us define longer prefixes!

Using hex feels a tiny bit weird, perhaps, because it requires a tiny bit of post-processing in a way that most of the schema-schema content doesn't, but I think it's not too out of line given the value provided.

rvagg:
mild cringe for feature creep for a feature that I don't see being used a whole lot outside of Filecoin which has a single byte, but putting the key in the place of a key would be nice
:shrug:

rvagg:
Changed based on a discussion with @warpfork:

* Now hexadecimal strings of arbitrary length
* DSL has them quoted, but I've made the `0x` prefix for the strings optional depending on user tastes (parsers can strip out the `0x`)
* DMT is now a map of `{String:TypeName}`, but the string here is still hexadecimal. So implementations using this would need to do the conversion and matching (e.g. Go codegen would probably want to do this conversion prior to generating and have the proper bytes handy)

warpfork:
I'm approximately 99% thrilled with this diff, other than being a little unsure on the "0x" thing.  Is that variance tolerance really significantly desired?  I feel like having an "0x" there adds a bit of complexity (not much, but also no real upside), and suggests we'll support alternatives that we probably don't.

warpfork:
We may also want to specify capitalization if we're using hex.  (I lean towards allcaps, though it's an incredibly weakly held position.)

If we do support variation in either of these things, I think we might best drop a comment in the spec now about which is canonical, so we don't have to revisit this when we write autoformatter tools.

warpfork:
Just to think through this outloud: by making `type UnionRepresentation_BytePrefix` into a map directly, rather than a struct which contains the same map, we do get to elide a little bit of depth in the schema-schema DMT... but we also lose the ability to add more optional fields to the struct in the future.

I don't actually have any concretely idea of something we might add to this, so it may well be a moot point, but it is a relatively irreversible choice.

... Yeah, I think we're fine here.  After all, the utterly "worst case" scenario to regrow something with more fields would be just adding another member to the union.  That's fine.

specs/schemas/schema-schema.ipldsch
```diff
@@ -425,15 +425,22 @@ type UnionRepresentation_Inline struct {
...
+## they should not be present in the internal representation, only valid
+## hexadecimal strings are allowed.
+##
 ## byteprefix is an invalid representation for any union that contains a type
 ## that does not have a bytes representation.
 ##
-type UnionRepresentation_BytePrefix struct {
-	discriminantTable {TypeName:Int}
-}
+type UnionRepresentation_BytePrefix {String:TypeName}
```
warpfork, line 443: We should probably change the name of this to `type UnionRepresentation_BytesPrefix` (with an 's') now, too.  :facepalm:

rvagg:
I did a s/byteprefix/bytesprefix, removed `0x` and clarified upper-case. ptal.

specs/schemas/schema-schema.ipldsch
```diff
@@ -423,17 +423,23 @@ type UnionRepresentation_Inline struct {
...
+## Only valid, upper-case, hexadecimal strings representing at least one byte
+## are allowed.
+##
+## bytesprefix is an invalid representation for any union that contains a type
 ## that does not have a bytes representation.
 ##
-type UnionRepresentation_BytePrefix struct {
-	discriminantTable {TypeName:Int}
-}
+type UnionRepresentation_BytesPrefix {String:TypeName}
```
warpfork, line 442: Just highlighting for the sake of clarity: this does defacto give up any possibility to add more features to UnionRepresentation_BytePrefix in the future.

But this is probably fine, because A) I don't have even speculative ideas today about what those "more features" could be, and B) worst case scenario, we can still add new members to UnionRepresentation union in the future, so we've got another flex point near to hand if we really need it.

warpfork (approved):
:ship:

PR #107, opened 2021-06-22
CARv2 Specification

rvagg:
(Builds on #101 which has the CARv1 spec promoted to the new docs site)

Rendered: https://github.com/ipld/ipld/blob/rvagg/carv2/specs/transport/car/carv2/index.md

This describes the existing format, currently described across a combination of:

* CARv2 serving as a DAG blockstore as described in: https://docs.google.com/document/d/118fJdf8HGK8tHYOCQfMVURW9gTF738D0weX-hbG-BvY (if you don't have access to this please ask, we can probably share it)
* https://github.com/ipld/specs/pull/248 which was an initial wishlist after we described CARv1 and understood its deficiencies, but mostly we're just using the pragma detailed in the discussion in this PR's comments
* Work going on in https://github.com/ipld/go-car/tree/wip/v2 to implement this spec
* https://github.com/willscott/carbs where the index specification comes from (specifically indexsorted.go)
* https://github.com/ipld/js-car/pull/30 has JavaScript read support, minus the index (I can share basic index parsing code if someone who speaks JS finds that helpful)

**This is a WORK IN PROGRESS and this PR is intended for discussion purposes so we can quickly pin down this format and move on to implementation.**

CARv2 isn't intended to be perfect, it's imagined that we could get closer to "perfect" with a v3. For now the priority is to use this as a relatively efficient blockstore with some extensibility to support some nice-to-haves.

I've been careful to describe what we have now, with the assumption that we may adjust these based on feedback (e.g. the signed vs unsigned in the index descriptor I'm imagining changing). Although I've done a lot of hand-waving in the characteristics field which I don't believe anyone has yet described how it's supposed to work other than a few bullet-points of what it could contain.


specs/transport/car/carv2/index.md
```diff
@@ -0,0 +1,177 @@
...
+CARv2 is a minimal upgrade to the [CARv1](../carv1/index.md) format with the primary aim of adding an optional index within the format for fast random-access to blocks.
+
+CARv2 makes use of CARv1 by wrapping a properly formed CARv1 with a prefix containing a pragma and header, and a suffix containing the optional index data. Once the offset and length of the CARv1 bytes are determined using CARv2 parsing rules, an existing CARv1 decoder can be used to read the roots and `CID:Bytes` pairs. Likewise, a CARv1 encoder can be used to encode this data for wrapping by a CARv2 encoder.
+
+## Format Description
+
+CARv2 consists of:
+
+1. An 11-byte pragma (or sequence of "magic bytes") that identify the data as a CARv2 format.
+2. A header describing data payload offset and length and index offset within the container as well as a bitfield for setting various capabilities supported by the container.
```
mvdan, line 34: should we list all that the header contains here? IMO it would be easier to simply refer to the Header section below.

specs/transport/car/carv2/index.md
```diff
@@ -0,0 +1,177 @@
...
+3. A standard CARv1 data payload, including standard CARv1 header and roots and sequence of `CID:Bytes` pairs.
+4. An optional index payload, which may be one of a number of supported index formats, allowing for fast lookups of blocks within the data payload in the form of an offset from the start of the container.
```
mvdan, line 36: I feel similarly about these two sections. I'd personally make them shorter and just link to the sections with details.

specs/transport/car/carv2/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+Existing CARv1 parsers should safely read this pragma and reject the container as an unsupported version of the CAR format.
+
+This 11 byte string remains fixed and may be matched using a simple byte comparison and does not require a varint or CBOR decode since it does not vary for CARv2 containers.
+
+### Header
+
+Following the 11 byte pragma, the CARv2 is a fixed-length sequence of 40 bytes, broken into the following sections:
+
+1. **Characteristics**: A 128-bit (16-byte) bitfield used to describe certain features of the enclosed data.
```
mvdan, line 85: Do we want to clarify that this will begin as all zeros, given that we're not using it yet?

This is important because, if we add any characteristic bits in the future, we have to keep in mind that all past CARv2 blobs will have the bits unset, for backwards compatibility.

specs/transport/car/carv2/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+The CARv2 format can be illustrated as follows:
+
+```
+| 11-byte fixed pragma | 40-byte header | optional padding | CARv1 data payload | optional padding | optional index payload
+```
+
+### Pragma
+
+The CARv2 version pragma (or "magic bytes") was chosen for compatibility with existing CARv1 parsers. CARv1 leads with a [DAG-CBOR](../../../codecs/dag-cbor/index.md) block, prefixed with a varint where the block contains the version and roots array:
```
masih, line 46: Being picky: would it make sense to eliminate all references to "magic", in favour of pure unadulterated logic? Would it add anything to introduce "magic bytes" as an alternative name for the pragma?

specs/transport/car/carv2/index.md
```diff
@@ -0,0 +1,177 @@
...
+This 11 byte string remains fixed and may be matched using a simple byte comparison and does not require a varint or CBOR decode since it does not vary for CARv2 containers.
+
+### Header
+
+Following the 11 byte pragma, the CARv2 is a fixed-length sequence of 40 bytes, broken into the following sections:
+
+1. **Characteristics**: A 128-bit (16-byte) bitfield used to describe certain features of the enclosed data.
+2. **Data offset**: A 64-bit (8-byte) unsigned little-endian integer indicating the byte-offset from the beginning of the container to the first byte of the CARv1 data payload.
+3. **Data length**: A 64-bit (8-byte) unsigned little-endian integer indicating the byte-length of the CARv1 data payload.
+4. **Index offset**: A 64-bit (8-byte) unsigned little-endian integer indicating the byte-offset from the beginning of the container to the first byte of the index payload.
```
mvdan, line 88: I'd maybe make a note here that this offset can be 0; none of the others can be 0, presumably. You do explain the 0 case further below, but I think it's important to note here too.

specs/transport/car/carv2/index.md
```diff
@@ -0,0 +1,177 @@
...
+  * [Index format](#index-format)
+    * [Format `0x300000`: IndexSorted](#format-0x300000-indexsorted)
+* [Implementations](#implementations)
+* [Test Fixtures](#test-fixtures)
+
+## Summary
+
+CARv2 is a minimal upgrade to the [CARv1](../carv1/index.md) format with the primary aim of adding an optional index within the format for fast random-access to blocks.
+
+CARv2 makes use of CARv1 by wrapping a properly formed CARv1 with a prefix containing a pragma and header, and a suffix containing the optional index data. Once the offset and length of the CARv1 bytes are determined using CARv2 parsing rules, an existing CARv1 decoder can be used to read the roots and `CID:Bytes` pairs. Likewise, a CARv1 encoder can be used to encode this data for wrapping by a CARv2 encoder.
```
masih, line 27: > Likewise, a CARv1 encoder can be used to encode this data for wrapping by a CARv2 encoder.

Thinking out loud: Though accurate, we probably want to encourage folks to use a CAR v2 encoder so that index can be generated as blocks are added. Otherwise, an archive produced by a CAR v1 encoder and passed to a CAR v2 for encoding would mean that blocks need to be re-scanned for the index to be generated.

specs/transport/car/carv2/index.md
```diff
@@ -0,0 +1,177 @@
...
+   * `3` indicates that the ordering is dictated by an [IPLD Selector](../specs/selectors/index.md) which itself is serialized into the CARv1 payload and referenced in the last element in the "roots" array. i.e. deserializing the block referenced by the last element of the roots array in the CARv1 header should yield a data model form of the IPLD selector that was used to select and order the blocks in the CARv1 payload.
+   * `4` has the same meaning as `3` except that the IPLD selector is encoded as a DAG-CBOR IPLD block immediately following the Header and is prefixed with a varint indicating the length of the bytes to decode.
+2. **Bit 8** indicating whether the blocks have been properly deduplicated (`1`) and that a reader should not expect any repeated CIDs, or that the blocks _may_ contain duplicates (`0`) (perhaps as a trade-off for speed or memory consumption during generation).
+
+Future amendments of this specification may introduce additional characteristics, including:
+
+* **TODO:** what else are we potentially filling 128-bits up with?
```
mvdan, line 106: I thought we were leaving the characteristics bits unused for now :) Your two ideas sound interesting here, though I wonder if we need them right now in the spec, given that the first version of go-car v2 won't support them.

specs/transport/car/carv2/index.md
```diff
@@ -0,0 +1,177 @@
...
+```
+a1                                                # map(1)
+  67                                              #   string(7)
+    76657273696f6e                                #     "version"
+  02                                              #   uint(2)
+```
+
+Existing CARv1 parsers should safely read this pragma and reject the container as an unsupported version of the CAR format.
+
+This 11 byte string remains fixed and may be matched using a simple byte comparison and does not require a varint or CBOR decode since it does not vary for CARv2 containers.
```
masih, line 79: > containers

Would the word "archives" be a better fit?

specs/transport/car/carv2/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+The CARv2 index payload follows the CARv1 data payload, but may be offset by padding or additional data as dictated by the characteristics field. The index payload begins at "Index offset", which _must_ be after the end of the CARv2 data payload, and continues until the end of the CARv2 byte stream (its length is not encoded in the header).
+
+An "Index offset" value of `0` in the CARv2 header indicates that there is no index in this CARv2 container and no attempt should be made to read it.
+
+### Index format
+
+The CARv2 index is a flexible container format itself, allowing for different index layouts depending on suitability for a particular application or set of data - generation speed, usage performance, size, etc. Supported index formats will be detailed in this specification, below.
+
+Index data, once read from the CARv2 container, provides a mapping of hash digest bytes to block location in byte offset from the beginning of the CARv2 container. The index only uses the hash digest&mdash;it does not use the full bytes of a CID, nor does it use any of the multihash prefix bytes.
```
mvdan, line 128: did you mean to write an HTML-escaped dash here? I think a `-`, `--`, or `;` would work in plaintext, but maybe I'm missing something.

specs/transport/car/carv2/index.md
```diff
@@ -0,0 +1,177 @@
...
+Existing CARv1 parsers should safely read this pragma and reject the container as an unsupported version of the CAR format.
+
+This 11 byte string remains fixed and may be matched using a simple byte comparison and does not require a varint or CBOR decode since it does not vary for CARv2 containers.
+
+### Header
+
+Following the 11 byte pragma, the CARv2 is a fixed-length sequence of 40 bytes, broken into the following sections:
+
+1. **Characteristics**: A 128-bit (16-byte) bitfield used to describe certain features of the enclosed data.
+2. **Data offset**: A 64-bit (8-byte) unsigned little-endian integer indicating the byte-offset from the beginning of the container to the first byte of the CARv1 data payload.
```
masih, line 86: "Data" sound fine. the wip branch of go-car v2 uses the phrase `CarV1`. Either works for me. I'd love to keep them consistent.

specs/transport/car/carv2/index.md
```diff
@@ -0,0 +1,177 @@
...
+3. **Data length**: A 64-bit (8-byte) unsigned little-endian integer indicating the byte-length of the CARv1 data payload.
```
masih, line 87: Similar to the previous comment, the phrase used in wip of go-car v2 is "Size". Both seem fine; I'd love to keep them consistent.

specs/transport/car/carv2/index.md
```diff
@@ -0,0 +1,177 @@
...
+4. **Index offset**: A 64-bit (8-byte) unsigned little-endian integer indicating the byte-offset from the beginning of the container to the first byte of the index payload.
+
+### Characteristics
+
+The characteristics bitfield contained within the CARv2 header may be used to indicate certain features of the specific CARv2 container.
```
masih, line 92: If we are to merge this spec to the main branch, would it suffice to simply state, `Characteristics` is reserved for future use? I wonder if we don't have to decide this now, should we at all.

specs/transport/car/carv2/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+The first byte(s) of a CARv2 index are a varint that indicates the index format type. The remaining bytes follow the encoding rules of that index format type.
+
+As the index only contains the hash digest bytes, other details contained within the block's CID and the length of the block's bytes must be derived by inspecting the initial bytes of the block entry within the data payload.
+
+**TODO:** How are we handling identity hash CIDs?
+* Do we want to state that identity multihash CIDs are explicitly excluded from indexes? 
+* Can we defer an assumption that a block store interface would return the bytes without having to look them up when it sees an identity CID? 
+* What are the implications of having an index that doesn't contain all of the blocks within the CAR?
+* Or do we want to explicitly exclude index CIDs entirely from the CARv2 format since they are arguably a waste of space?
```
mvdan, line 138: I think it makes sense to say that one shouldn't use CAR files for identity hash CIDs. In general, it just doesn't make sense.

I also think it's reasonable for the index to refuse to index those; that allows the index to know that hashes are relatively small, for example. If we had an API that takes a CID or Hash and gives you an `io.Reader` of the data in Go, the logic could be like:

1) Is it an identity hash? If so, simply return a reader of the identity hash itself
2) Otherwise, look it up in the index, and return a reader of the block slice in the CARv1

Something worth thinking about is the ability to store entire DAGs in a CAR file. If we forbid storing identity hash CIDs as blocks in this spec, does that mean it's impossible to store a complete DAG in a CARv2 when it has an identity hash CID link? All the data would be there, since the link would contain the whole block, but strictly speaking some CIDs would be missing from the CARv1.

specs/transport/car/carv2/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+An index format type of `0x300000` indicates the "IndexSorted" format.
+
+IndexSorted sorts hash digests by two dimensions: first into buckets of _digest length_, smallest to largest, and then within those buckets ordered by a simple byte-wise sorting. In this way, locating a hash digest within the CAR requires first finding the bucket matching the length of the requested hash digest, then searching the ordered list of digests within that bucket to find the matching entry.
+
+* IndexSorted may contain one or more length-grouped buckets of digests.
+* Buckets are ordered by digest length and concatenated together to form the index.
+* Each bucket is prefixed with:
+  * a "width" encoded as a 32-bit **signed** little-endian integer indicating the common byte length of the combination of hash digests and their offsets in this bucket; followed by
+  * a "count" encoded as a 64-bit **signed** little-endian integer which determines the total number of hash digests (and their offsets) bucket.
```
mvdan, line 152: How come these two are signed? They could be unsigned from what I can tell, and then one doesn't have to worry about negative values.

specs/transport/car/carv2/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+* **TODO:** what else are we potentially filling 128-bits up with?
+
+### CARv1 data payload
+
+The CARv1 data payload starts at the "Data offset" indicated by the CARv2 header. It is important that a decoder adhere to this offset rather than assuming the data payload begins immediately after the header in order to make allowance for additional data following the header as indicated by the characteristics field in future amendments to this specification.
+
+The CARv1 data payload ends after the number of bytes indicated by the "Data length" field in the header. This is important as the CARv1 parser has no means to determine the end of the container other than encountering the end of a byte stream. Therefore, a CARv2 decoder may defer to a CARv1 decoder to load the `CID:Bytes` sequence payload and "roots" array, but must be able to deliver it a byte stream that strictly begins at "Data offset" and ends at "Data length".
+
+The CARv1 data payload must include a valid CARv1 header, including a roots array and a `version` field with a value strictly of `1`.
```
masih, line 114:
```suggestion
The data payload is a complete, self-contained CAR v1 archive. As such, it must include a valid CARv1 header, including a roots array and a `version` field with a value strictly of `1` followed by a series of data blocks.
```

I also wonder if we want to say a sentence about why. e.g. "This allows conversion of a CAR v2 to a CAR v1 without any further processing."

specs/transport/car/carv2/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+* IndexSorted may contain one or more length-grouped buckets of digests.
+* Buckets are ordered by digest length and concatenated together to form the index.
+* Each bucket is prefixed with:
+  * a "width" encoded as a 32-bit **signed** little-endian integer indicating the common byte length of the combination of hash digests and their offsets in this bucket; followed by
+  * a "count" encoded as a 64-bit **signed** little-endian integer which determines the total number of hash digests (and their offsets) bucket.
+
+A common case of a single bucket of 32-byte hash digests is expected due to the commonality of this digest length for CIDs.
+
+Individual index entries are the concatenation of the hash digest an an additional 64-bit **unsigned** little-endian integer indicating the offset of the block from the begining of the CARv2 container. Offsets locate the first byte of the varint that prefix the `CID:Bytes` pair within the CARv1 payload. See the [data section in the CARv1 Specification](../carv1/index.md#data) for details on block encoding.
```
mvdan, line 156: Thinking to myself for a second: if the indexed offsets are from the beginning of the CARv2 container, this means that changing the Data Offset (padding before CARv1) requires changing all the offsets in the index. I guess that won't be a common operation, but another option would be to make all these offsets relative to Data Offset, i.e. relative to the start of the CARv1.

specs/transport/car/carv2/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+The CARv1 data payload ends after the number of bytes indicated by the "Data length" field in the header. This is important as the CARv1 parser has no means to determine the end of the container other than encountering the end of a byte stream. Therefore, a CARv2 decoder may defer to a CARv1 decoder to load the `CID:Bytes` sequence payload and "roots" array, but must be able to deliver it a byte stream that strictly begins at "Data offset" and ends at "Data length".
+
+The CARv1 data payload must include a valid CARv1 header, including a roots array and a `version` field with a value strictly of `1`.
+
+Refer to the [CARv1 Specification](../carv1/index.md) for details on the CARv1 format.
+
+### Index payload
+
+The CARv2 index payload follows the CARv1 data payload, but may be offset by padding or additional data as dictated by the characteristics field. The index payload begins at "Index offset", which _must_ be after the end of the CARv2 data payload, and continues until the end of the CARv2 byte stream (its length is not encoded in the header).
```
masih, line 120: > which _must_ be after the end of the CARv2 data payload

**if the index is present is present.**

specs/transport/car/carv2/index.md
masih, line 128: > from the beginning of the CARv2 container.

Do we mean to say from the beginning of CAR v1 data payload? 
If not this is inconsistent with requirements listed [here](https://github.com/filecoin-project/dagstore/pull/2/files#diff-b0929dacd4f1fc192cd308c1a9db2e85154ea742ccec315518e659d7f6a3efc7R334-R335)

specs/transport/car/carv2/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+An "Index offset" value of `0` in the CARv2 header indicates that there is no index in this CARv2 container and no attempt should be made to read it.
+
+### Index format
+
+The CARv2 index is a flexible container format itself, allowing for different index layouts depending on suitability for a particular application or set of data - generation speed, usage performance, size, etc. Supported index formats will be detailed in this specification, below.
+
+Index data, once read from the CARv2 container, provides a mapping of hash digest bytes to block location in byte offset from the beginning of the CARv2 container. The index only uses the hash digest&mdash;it does not use the full bytes of a CID, nor does it use any of the multihash prefix bytes.
+
+The first byte(s) of a CARv2 index are a varint that indicates the index format type. The remaining bytes follow the encoding rules of that index format type.
```
masih, line 130: Being picky: [it's a `uvarint`](https://github.com/ipld/go-car/blob/wip/v2/v2/internal/index/index.go#L64).

specs/transport/car/carv2/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+As the index only contains the hash digest bytes, other details contained within the block's CID and the length of the block's bytes must be derived by inspecting the initial bytes of the block entry within the data payload.
```
masih, line 132: Is this true for all index types? i.e. [`mapGobIndex `](https://github.com/ipld/go-car/blob/wip/v2/v2/internal/index/indexgobhash.go) and [`mapIndex `](https://github.com/ipld/go-car/blob/wip/v2/v2/internal/index/indexhashed.go)

specs/transport/car/fixture/carv2-basic/carv2-basic.json
```diff
@@ -0,0 +1,114 @@
...
+      "cid": {
+        "/": "Qmcpz2FHJD7VAhg1fxFXdYJKePtkx1BsHuCrAgWVnaHMTE"
+      },
+      "content": {
+        "Links": [
+          {
+            "Hash": {
+              "/": "bafkreifuosuzujyf4i6psbneqtwg2fhplc2wxptc5euspa2gn3bwhnihfu"
+            },
+            "Name": "fishmonger",
```
masih, line 62: fishy naming 😊

masih (commented):
Nice write up 🍻

specs/transport/car/carv2/index.md
```diff
@@ -0,0 +1,177 @@
...
+The CARv2 version pragma (or "magic bytes") was chosen for compatibility with existing CARv1 parsers. CARv1 leads with a [DAG-CBOR](../../../codecs/dag-cbor/index.md) block, prefixed with a varint where the block contains the version and roots array:
+
+```ipldsch
+type CarV1Header struct {
+  version Int # 1
+  roots [&Any]
+}
+```
+
+To introduce a new version that existing parsers can safely reject as an *"unsupported version"* we must copy the minimal form of this header with a new version.
```
mikeal, line 55: Did we actually implement this in existing parsers? It would be good to know the failure modes we expect ahead of time, because throwing on version detection and missing a root are going to look very different.

specs/transport/car/carv2/index.md
```diff
@@ -0,0 +1,177 @@
...
+1. **Characteristics**: A 128-bit (16-byte) bitfield used to describe certain features of the enclosed data.
+2. **Data offset**: A 64-bit (8-byte) unsigned little-endian integer indicating the byte-offset from the beginning of the container to the first byte of the CARv1 data payload.
+3. **Data length**: A 64-bit (8-byte) unsigned little-endian integer indicating the byte-length of the CARv1 data payload.
+4. **Index offset**: A 64-bit (8-byte) unsigned little-endian integer indicating the byte-offset from the beginning of the container to the first byte of the index payload.
+
+### Characteristics
+
+The characteristics bitfield contained within the CARv2 header may be used to indicate certain features of the specific CARv2 container.
+
+**TODO:** spit-balling here, these feel like feature-bloat so we should probably only add features as we use them but should be clear to leave enough space to encode additional things we care about without causing conflicts - i.e. it shouldn't be possible to flip bits to say that it's _both_ depth-first and breadth-first. Hence the suggestion to use easy integers where there's >2 options.
```
mikeal, line 94: Why not make this a CID:Data entry of dag-cbor instead of a custom data format? The first block being the header would be fine, it would future proof this a lot better too.

specs/transport/car/carv2/index.md
mikeal: I agree, there’s really no utility in including them, they appear whole in any block that references them, that’s enough.

specs/transport/car/fixture/carv2-basic/index.md
```diff
@@ -0,0 +1,11 @@
...
+title: "carv2-basic"
+eleventyNavigation:
+  synopsys: "Basic CAR with linked raw and DAG-PB"
+---
+
+carv2-basic
+==========
+
+* [carv2-basic.car](./carv2-basic.car) - raw CAR
+* [carv2-basic.json](./carv2-basic.json) - description of the contents and layout of the raw CAR, encoded in DAG-JSON
```
warpfork, line 11: Big, big love for this directory and all of its contents.

specs/transport/car/carv2/index.md
rvagg: :thumbsup: I'll soften the language to make it more of a theoretical point:

>  Though not necessarily ideal, an existing CARv1 decoder could be used to read the roots and `CID:Bytes` pairs. Likewise, a CARv1 encoder could be be used to encode this data for wrapping by a CARv2 encoder as the payload is the same format.

specs/transport/car/carv2/index.md
rvagg: :thumbsup: have tried to shorten a little, if you see opportunities for more then let me know. I want to get in the little code block below and these points give enough context for it not to be too much new information by the time you see it.

specs/transport/car/carv2/index.md
rvagg: There's no such thing as pure unadulterated logic!

But also, I was thinking of linking to https://en.wikipedia.org/wiki/List_of_file_signatures (which also calls them "Magic Bytes") to help make the point that this isn't a huge number of bytes for this purpose compared to other formats.

I've removed the first reference in the description above but left this one and added the link. Let me know if it still smells to you, I'm happy to remove as it's not a big deal.

specs/transport/car/carv2/index.md
rvagg: Aye:

* https://github.com/ipld/js-car/blob/da397a8b3659a3e3ead9411746962885537ec1ac/test/test-errors.js#L38-L40
* https://github.com/ipld/go-car/blob/71cfa2fc2a619d646606373c5946282934270bd4/car_test.go#L247-L250

specs/transport/car/carv2/index.md
rvagg: Not really, cause it'd be "Content Addressable aRchives archives". I could just try stripping out the descriptor though. I'm consciously avoiding "file" here, same as the v1 spec, which is also why this spec is in a section called "transports". In the v1 I didn't use "container" and stuck to either just saying "CAR" or "CAR format".

I've removed all of the uses of "container" here, I'm not sure if it's an improvement, however, it makes heavy use of "CARv2" as a noun now. Let me know what you think.

specs/transport/car/carv2/index.md
rvagg: :thumbsup:

> All bits in the bitfield will be unset (`0`) by default and only set (`1`) where they are being used to signal a characteristic other than the default.

specs/transport/car/carv2/index.md
rvagg: hm, I went with "size" in JS too but "length" here because so many ways we talk about bytes are in "length". I've tried just renaming "Data length" to "Data size" but left the other "length"s alone. Not sure if it's an improvement but probably not a big deal.

specs/transport/car/carv2/index.md
rvagg: :thumbsup:

> This value may be `0` to indicate the absence of index data.

specs/transport/car/carv2/index.md
rvagg: I think I'd prefer "data" on this one, I don't think the discrepancy with code is a big deal though, on this one at least.

specs/transport/car/carv2/index.md
rvagg: I mainly ask this because 128-bits is a lot of bits and I'd like to see what's behind wanting so much optionality here. This TODO is under "Future ammendments ... including:", if we can't come up with many ideas then maybe we need to do some careful thinking about this field.

specs/transport/car/carv2/index.md
rvagg: Yeah, I'm not really excited about fact that we're yet again inventing another custom format when we have good formats at hand already.

Maybe the main reason to choose this was has to do with determinism of location, which is a problem we have in CARv1 now since the header is a flexible size and coming back and updating it after you've written the data is difficult.

But it's true that these fields are all fixed-length and don't have the problem of containing a flexible-length CIDs array like the CARv1 header so this could easily be just a CBOR array. If we're future proofing we'd need to add a varint length prefix too so we know how big the header is. I don't know if we need the CID, CARv1 doesn't bother doing that with the header block (which does add inconsistency to the format which is slightly annoying, but it's precedent).

In IPLD Schema it could be:

```
type CARv2Header struct {
  characteristics Bytes
  dataOffset Int
  dataSize Int
  indexOffset Int
} representation tuple
```

specs/transport/car/carv2/index.md
rvagg: :thumbsup:

specs/transport/car/carv2/index.md
rvagg: Yes, I did, cause I was doing it on Linux where I can never remember the keyboard shortcut for em-dash, so this is how I insert them in that situation. (I'm a stickler for proper em-dashes and their usage). Have fixed now.

rvagg: Oh! Interesting. I hadn't actually picked up on that, but it makes sense (and I notice I say this is the case later in the doc!). Thanks.

specs/transport/car/carv2/index.md
rvagg: it's only a `uvarint` in Go

I've updated it to be clearer, with the long-form description of it being an unsigned LEB128 integer that I used in the CARv1 spec.

specs/transport/car/carv2/index.md
rvagg: :shrug: tbh I hadn't looked at the other types but did wonder if I was being too narrow here.

My thinking though is that we're only doing a single index type for this initial version, `IndexSorted`, we can be somewhat narrow here but open it up more as we need descriptive flexibility for other formats.

I don't think we're planning on shipping or specing any other index formats right now, are we? That seems like overreach and we should go with the most basic and performant format for now and only add as we have a provable need to expand. @willscott - input on this since CARBs is your baby?

specs/transport/car/carv2/index.md
rvagg: OK, so next question - do we take this opportunity to explicitly rule them out of the CARv1 contained within the CARv2 here? Or is that doing narrowing down the existing spec a bit too much? This will end up being a question of determinism for cases where it matters (like Filecoin deals). Maybe this is a "characteristic"?

specs/transport/car/carv2/index.md
rvagg: cause they are signed currently, I bolded them to highlight this fact and assumed we'd get agreement to change that: https://github.com/ipld/go-car/blob/2a49cbd958b0c6443dd1e9b518814199bf50842d/v2/internal/index/indexsorted.go#L21-L22

@masih maybe let me know when they're changed and I'll update here, I just don't want to risk the spec & impl being out of sync.

specs/transport/car/carv2/index.md
rvagg: Thanks to @masih for pointing out that the intention was to be offsets from the beginning of the CARv1, so this is wrong. It ought to be possible to hold the index and CARv1 in separate files and for them still to make sense without modification.

specs/transport/car/carv2/index.md
rvagg: no, before merging I'll clean this up with the state of things as they are - I'm really just trying to get some discussion / visibility on this topic as it feels too much to me like we're handwaving some hypotehtical future needs here without having a clear enough vision. Either we need this or we don't or maybe we should be affording ourselves a little more flexibility - as noted below in the dag-cbor discussion.

specs/transport/car/carv2/index.md
rvagg:
```
type CARv2Header struct {
  dataOffset Int
  dataSize Int
  indexOffset Int
  characteristics Bytes
} representation tuple
```

With `characteristics` last might make more sense. But as long as we're going with dag-cbor we really should consider this field and whether we should just add explicit entries for it. `deduped Bool`, `orderingMethod Int`.

rvagg:
k, pushed a big update addressing most of the feedback so far. Happy to receive more input.

Some outstanding items:

1. Can we switch to a `<varint>dagcbor([dataOffset, dataSize, indexOffset, characteristics])` style header block instead of the custom layout we have now - inventing new encoding schemes is something we have a habit of doing while we have perfectly good, stable and deterministic formats that are well tested an implemented right at our hands. The big reason why this was a problem for CARv1 is the flexibility in length of the `roots` array which makes it very hard to rewind and update - plus the fact that it's a map and not an array which puts `version` last. But we don't have flexible-length items here. Using DAG-CBOR would actually give us a lot of flexibility to do interesting things with the header into the future and we could even get rid of the characteristics field entirely for this first pass since we're not using it, but it would be easy to add back in later -- or even better, just add in single fields for each characteristic we're adding.

2. Do we want to rule out identity CIDs from the CARv1 contents - it's now ruled out of indexes as per feedback, but I don't believe we have anything in our existing implementations that would stop them from being stored in CARv1s today. Maybe this is just something we punt for when we have flags regarding determinism?

3. Are we shipping more than one index type with our initial version? My assumption was that we're shipping the most basic and sensible index type and would ship more later as we decided we needed them for `<reasons>` (e.g. perf). If we're doing >1, I need to understand and spec them here.

4. Getting those signed ints to be unsigned inside the go-car work.

masih:
> Can we switch to a `<varint>dagcbor([dataOffset, dataSize, indexOffset, characteristics])` style header block instead of the custom layout we have now 

There must be a good answer to this but why can't we do this with the current car v1 header since it is CBOR already? what is it about that second CBOR that fixes broken things in the CBOR block that comes before it?

mvdan:
> @rvagg in the context of pragma for v2 would you mind reminding me why not having `roots` in the header doesn't cause parse failure for v1 clients but adding extra fields would?

The CARv1 header is dag-cbor, and part of the determinism is that we're strict about what fields are optional or not. The spec defines both struct fields (version and roots) as non-optional, so both must be present.

mvdan:
> There must be a good answer to this but why can't we do this with the current car v1 header since it is CBOR already?

Adding extra fields/keys to the CARv1 header encoded as a dag-cbor map would break existing decoders. dag-cbor being strict, it will error on unknown keys: https://github.com/ipld/go-ipld-prime/blob/e798ae8c2e5a96f3d5a184e4a236c843eac0b1ec/codec/dagcbor/unmarshal.go#L96-L99

Having to go back and teach carv1 decoders about the new fields to be ignored beats backwards compatibility :)

masih:
"Dag-CBOR" is what I was missing; much obliged @mvdan 🍻

mvdan:
On point 1: It's taken me some time to catch up with @rvagg, but I finally agree with him on making the CARv2 header a dag-cbor tuple. I also agree we should remove Characteristics and just add more fields at the end of the tuple later on.

2: I don't think we can outright forbid identity CIDs from the inner CARv1, unless the CARv1 spec already forbids them. Otherwise, wrapping a CARv1 in a CARv2 would also involve iterating over the entire CARv1 to remove identity CID entries - that could get quite expensive with large CAR files. I think it should be opt-in at a later time, like "make CARv1 entries sorted and deterministic", because that would also require a "wrap CARv1 with a CARv2" to sanitise the CARv1.

3: The scope of this initial CARv2 version is what you can see :) So no more index formats for now.

specs/transport/car/carv2/index.md
masih: Thanks for pointing this out @rvagg

masih: This pr should fix it https://github.com/ipld/go-car/pull/96

willscott:
> Can we switch to a <varint>dagcbor([dataOffset, dataSize, indexOffset, characteristics]) style header block

I think we're going to end up being unhappy with this given the known workflow of attaching and detaching the index.
We really want the header to be a fixed length so that the offset of the carv1 in the file doesn't change as we change the datasize and indexoffset values.

mvdan:
We could possibly use a large enough dataOffset (byte offset for the inner carv1) so that we could encode any current dagcbor header without having to increase the offset, but that gets tricky. Like Will said on slack, what if increasing dataOffset makes its varint be larger, so we then need to increase dataOffset again? It's a bit of a loop. We also can't tell if we would add more elements to the array later on, so it's really hard to tell what offset would be large enough.

So I now agree with Will that the header should be fixed size :) We can leave the characteristics bits in there, to make our life easier in the future for adding boolean options. We haven't defined how we would add any extra data that doesn't fit in characteristics bits, though we also have padding between the carv1 and the index, so I think the idea was that in the future we could add extras there without breaking any carv2 decoders.

rvagg:
> what if increasing dataOffset makes its varint be larger

So this isn't the problem, no varint involved here other than to say how long the dag-cbor block is, which is stable in my initial conception above as long as we're not adding fields, and you shouldn't be rewriting the header with new fields after it's written I think. So write header, then just come back and write a new one with the same fields, same size because you don't have the flexi-roots problem.

But I woke up at 3am night last night and remembered the smallest-representation-size rule we have as a strict requirement in DAG-CBOR for integers! (don't take your work to bed folks). So if you wrote your first pass header with offsets of `0` and came back and bumped them up (iirc even past `23` you overflow to a new byte in CBOR). They're not varints, but they're smallest-size ints - 8, 16, 32, 64 batches, with a special case for <=23. So that sucks!

But it's not impossible to deal with, you could just mandate that the space for the header block occupies the maximal sized form of it - but yeah .. sloppy.

On the other hand we have a flexibility problem that we're _sort of_ solving for with this "characteristics" field which we have no use for yet but are guessing that 128-bits will be golden for future uses. I think that's a problem and we're either overengineering by having it, or not being prepared for potentialities because we can't imagine them but we're sure we're going to need _something_.

So, how about this for a proposal: **drop characteristics entirely from this initial version**, leaves only 3 uint64's for offsets following the pragma. But, because we have the offset for the carv1 payload, we still have the ability to put it back in later, as an optional extra, in a way that even the initial code that supports CARv1 parsing can read (it'll just ignore it and consider it padding). But, it also gives us flexibility to do more useful things with this little beasty. Perhaps it's not a bitfield, but a set of uint8's, or perhaps we want to throw a selector in there as well so we just make the thing that follows the 3 uint64's a `varint+dag-cbor()` block with all the fields we care about.

specs/transport/car/carv2/index.md
rvagg: doc updated

specs/transport/car/carv2/index.md
masih: I'll change the code to reflect this

specs/transport/car/carv2/index.md
masih: no big deal indeed thanks @rvagg

specs/transport/car/carv2/index.md
masih: That makes sense thank you

specs/transport/car/carv2/index.md
masih: This is perfectly fine; please ignore me 😊

rvagg:
k, discussions about header format got a bit off track and received enough pushback that we're just going to roll with what's here now.

I've addressed outstanding TODOs, I've also started a multicodec registration for `IndexSorted`, it'll probably be `0x0400`. @masih: that `0x300000` [here](https://github.com/ipld/go-car/blob/wip-v2/v2/index/index.go#L17) will need to be updated to `0x0400` - until we register further multicodecs and list the index formats here we should probably remove them from the mapping table. If we need them we can get them into here and register a multicodec relatively quickly.

would appreciate eyes again and signoff if others think this is good to go as a first round of the spec - it doesn't need to be perfect and we can iterate (it's listed as "Draft" and only needs to match what we're actually doing now).

rvagg:
Rendered, if it helps with reviews: https://github.com/ipld/ipld/blob/rvagg/carv2/specs/transport/car/carv2/index.md

mvdan (approved):
SGTM as a draft. I will be updating go-car's wip-v2 to fix the index multicodecs.

I left a reply to your question about identity CIDs, not sure if you saw. TL;DR I think the spec could clarify that the index does not contain identity hash CIDs. If we also want to skip them from the inner carv1, that would be a characteristic bit just like "deterministically sorted".

rvagg:
@mvdan :thumbsup: I did clean up the clarification in the latest version re identity CIDs:

> Indexes **should not include identity hash CIDs**. It is assumed that any use of a CARv2 as a blockstore will return identity CID data immediately by extracting it from the CID, therefore there should be no need to provide indexing for such entries.

PR #111, opened 2021-06-28
Tendermint and Cosmos specs

i-norden:
This is a port of https://github.com/ipld/specs/pull/369 to this repository.

WIP codecs can be found here: https://github.com/vulcanize/go-codec-dagcosmos


rvagg:
Build failure:

```
2022-04-21T13:29:13.736Z: ./specs/codecs/dag-cosmos/cosmos_state.md
2022-04-21T13:29:13.736Z: error: bad link specs/codecs/dag-cosmos/cosmos_state/typed_protobuf at line 11
2022-04-21T13:29:13.736Z: error: bad link specs/codecs/dag-cosmos/cosmos_state/typed_protobuf at line 36
```

Hopefully putting `.md` in there fixes that, although I suspect it might also be a problem because they're designed to be .html so you may have to resort to making a directory called `typed_protobuf` and putting an index.md in there.. I'd like to be able to avoid this kind of crazy but currently our SSB doesn't play nicely with the .md -> .html distinction.

specs/codecs/dag-cosmos/crypto_types.md
```diff
@@ -0,0 +1,29 @@
...
+# exclude the root hash.
+# This convention is implemented across IAVL range proofs as well.
+# Keep this consistent unless there's a very good reason to change
+# everything.  This also affects the generalized proof system as
+# well.
+type Proof struct {
+	Total    Int  # Total number of items.
+	Index    Int # Index of item to prove.
+	LeafHash Hash # Hash of item value.
+	Aunts    Aunts # Hashes from leaf's sibling to a root's child.
```
rvagg, line 27: Is there any utility in _also_ presenting an `AuntsLinks` or `AuntsCIDs` in the data model where you reinterpret the list of hashes as CIDs that can be navigable? Or is this purely a useful construct for the proof itself and navigation isn't really helpful at this point?

specs/codecs/dag-cosmos/basic_types.md
```diff
@@ -0,0 +1,39 @@
+# Tendermint and Cosmos Data Structure Basic Types
+
+```ipldsch
+# Uint is a non-negative integer
+type Uint bytes
```
rvagg, line 5: should this be an `int`? or is there something special about the representation?

```suggestion
type Uint int
```

specs/codecs/dag-cosmos/tendermint_chain.md
```diff
@@ -0,0 +1,355 @@
...
+
+#Txs is a list of Tx.
+type Txs [Tx]
+
+# TxCID is a link to a TX
+# TxCID uses the SHA256 multihash and the Raw codec (0x55)
+type TxCID &Tx
+
+# Tx is an arbitrary byte array.
+# Tx has no types at the Tendermint level, instead it encodes arbitrary length-prefixed data that can be decoded and processed in a state-machine specific manner.
```
rvagg, line 195: ! this is pretty frustrating .. I guess this is one of the points where you need the adjacent data structures just to piece the data together again into a consistent graph

specs/codecs/dag-cosmos/tendermint_chain.md
```diff
@@ -0,0 +1,355 @@
...
+
+## Tendermint Commit
+```ipldsch
+# CommitTreeCID is a CID link to the root node of a Commit merkle tree
+# This CID is composed of the SHA_256 multihash of the root node in a Commit merkle tree and the CommitTree codec (tbd)
+# Commit merkle tree is a Merkle tree built from the Signatures in a block's LastCommit
+type CommitTreeCID &MerkleTreeNode
+
+# Commit contains the evidence that a block was committed by a set of validators
+NOTE: Commit is empty for height 1, but never nil.
```
rvagg, line 207:
```suggestion
# NOTE: Commit is empty for height 1, but never nil.
```

specs/codecs/dag-cosmos/tendermint_chain.md
```diff
@@ -0,0 +1,355 @@
...
+# Value union type used to handle the different values stored in leaf nodes in the different merkle trees
+type Value union {
+    | SimpleValidator "validator"
+    | Evidence "evidence"
+    | TxCID "tx"
+    | Part "part"
+    | ResponseDeliverTx "result"
+    | Bytes "header"
+    | CommitSig "commit"
+} representation keyed
```
rvagg, line 99: Are these keyed unions in the data itself, or is it a mechanism to produce usable data structures? I note the same use of `&MerkleTreeNode` throughout but then dealing with the differentiation at the `keyed` representation level of `Value`, but do these keys actually exist? Is the alternative of having separate `SimpleValidatorMerkleTreeNode` and friends just too verbose for all of these different types.
My guess is that these keys aren't in the source data, and that if you traverse through a `&MerkleTreeNode` by reading the raw data that you'll end up in a place where you'll have to have retained information about which data structure you're in in order to read the next bit. Or, is this one of the data structures that you're using for accounting purposes and doesn't relate directly to chain data?

i-norden: Yeah that's correct, these keys don't actually exist in the source data and you'd need to retain information about which data structure you're in in order to read the next bit. Thanks for catching this, I think what we want to do is similar to how this is represented/handled for the various ethereum MMPTs whereby we have this [generic definition](https://github.com/ipld/ipld/blob/master/specs/codecs/dag-eth/state.md#trie-node-ipld) for `MerkleTreeNode`, including the leaf node with this `Value` union, but then we explicitly define specific subtypes of the `MerkleTreeNode` for each type of tree (per value) e.g. how we have [TxTrie](https://github.com/ipld/ipld/blob/master/specs/codecs/dag-eth/state.md#transaction-trie-ipld), [RctTrie](https://github.com/ipld/ipld/blob/master/specs/codecs/dag-eth/state.md#receipt-trie-ipld), etc, for ethereum. And these will each have their own multicodec byte prefix that specifies which value is stored in that particular tree and maps to a codec that can handle the encoding/decoding of the values in that tree.

specs/codecs/dag-cosmos/basic_types.md
i-norden: This should 100% be int, thanks!

specs/codecs/dag-cosmos/tendermint_chain.md
i-norden: Yes I am really struggling with how to represent this more rich type information. These txs map to a single cosmos state machine, the state machine which can make sense of these arbitrary bytes. So we could create a "namespacing" system for different Tendermint/Cosmos blockchains on IPLD, and use that to map these txs to the state machine that processes them. I'm not sure if that would involve registering a multicodec type for every Cosmos blockchain, or if we could take an approach similar to the one proposed here for supporting arbitrary protobuf objects where we prepend these bytes with a hash reference to an object type that provides that namespacing.

specs/codecs/dag-cosmos/crypto_types.md
i-norden: I think it would be valuable to make it navigable, will make this update.

specs/codecs/dag-cosmos/tendermint_chain.md
i-norden: Hmm actually I retract that statement, as I don't think we achieve any better typing than we already have. The `Header` IPLD has `ChainID` and `Version` fields, and these two fields specify the type of state machine and its specific version that is capable of making sense of the tx bytes.

i-norden: Aaand I retract my retraction, if we assume CosmosSDK standards are followed- which we already do elsewhere- then these tx bytes are always serialized protobuf messages and we can provide richer typing using the same `TypedProtobuf` scheme we propose here for typing arbitrary values stored in IAVL or SMT state. Updated the spec to reflect this, as well as cleaning things up some more.

rvagg:
@i-norden let's get this merged, we just have build failures, see the fleek-ci action here (or `make build` should do it locally I think) https://github.com/ipld/ipld/pull/111/checks?check_run_id=6245808898. I covered one of the failures @ https://github.com/ipld/ipld/pull/111#issuecomment-1106020161 but it looks like we have a couple more than that.

i-norden:
@rvagg sorry for all the `push -f`s and squashes, I couldn't seem to produce the errors with local `make build`. Not sure why but the fix appears to be to use `../` instead of `./`.

rvagg:
not a problem, we have a green so let's :ship:

PR #112, opened 2021-07-02
Start a doc page for collecting reference material about encryption.

warpfork:
This is just a quick start, probably much more to come -- and thanks to @oed for handing me the links that primed this content.

### DIFF
--- a/docs/synthesis/encryption/index.md
+++ b/docs/synthesis/encryption/index.md
@@ -1,0 +1,53 @@
+---
+title: Working With Encryption
+navtitle: Encryption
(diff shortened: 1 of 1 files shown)


warpfork:
I think it might be desirable to try to structure this oriented around solutions, and group them into the service they provide (authn vs authz, etc) -- but that's gonna need some hammering out, so I may just move to merge these in this current ultrabrief form that just hands off links to the other people working in the space, and leave readers to navigate that as they like.

warpfork:
Gonna merge this -- better to start gathering a few links and then iterate :)

PR #115, opened 2021-07-07
Add 'alice-words' fixture for HashMap spec

rvagg:
I still can't compile the site locally as mentioned in #114 but 🤞 this is good.


rvagg:
Prematurely hit "Comment" .. take 2:

I finally got my JS implementation using the new block layout so I can produce data that _should_ match the spec. I'm also doing Schema validation of block saves and loads so I know I'm confirming to the block form spec at least and I'm pretty confident all the rest is solid as it's heavily tested and has been put through a bit of practical of use already. But it'd be interesting to try and get the Go impl matching this.

https://github.com/rvagg/js-ipld-hashmap/pull/3 - has alice-words in the tests and also a way to generate the CAR and verify from a CAR.

mvdan:
Late due to OOO but SGTM :)

PR #116, opened 2021-07-12
links to schema spec content

warpfork:
Crosslink schema docs and schema specs, and actually link schema specs content.

It's necessary to do the schema specs content linking manually,
because the eleventy autogen stuff we've been using for other kinds of
content only works for the markdown content pages (or stuff with
frontmatter, perhaps, but that still excludes these files).

Spartan, but makes the content reachable.  Previously it was not.


warpfork:
Would like to prettify this more, but to have the linking present at all is an improvement, so I'm going to merge this without delay (as soon as CI confirms), and welcome iterative improvements on top of it later.

PR #117, opened 2021-07-13
clean up child listings

rvagg:
I originally wanted to do this just to get vertical alignment in the table fixed, but while in there I swapped the `font-size: 120%` for a `font-weight: bold` and IMO it's much more pleasant on the eyes. It also removes the "no synopsys available" since we have very few synopses in the entire site so this shows up far too much.

Thoughts?

Screenshot 2021-07-13 at 10 56 10 am


rvagg:
ping @warpfork, just a visual tweak and removing default text

PR #119, opened 2021-07-13
Add "schema-schema" make target

rvagg:
We shouldn't have to touch schema-schema.ipldsch.json _unless_ we're making fundamental changes that break the JS parser.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,10 @@ build:
 	@echo ">>> $@ >>>"
 	npm run build
 
+schema-schema:
+	@ echo ">>> $@ >>>"
+	npm run build:schema-schema
+
 dev:
 	@echo ">>> $@ >>>"
 	npm run dev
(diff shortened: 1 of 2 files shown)


warpfork:
That incantation terrifies me a little bit.  `npx` but we didn't have to add anything to any lockfiles?

Will it run on every build?  (What does the ":" do in npm build config, is that a rune of power?)

rvagg:
`:` is just a convention. Nothing automatic about this, entirely manually invoked by whoever wants to update schema-schema.
In #120, for example, this doesn't even work because it changes the schema DSL and the parser borks on it. So it'd be annoying to make it automatic and required.
`npx` to run it straight from npm, latest version. We can be specific with versions (`@version` after package name), but we also control this package entirely so I don't see a reason to be overly constrained. If it were run on every build then I'd put it in package.json and the lockfile.

warpfork:
I feel a tad dodgy about how this is glued together, but if it's gonna be a development productivity boost... okay, fine.  We can polish it better later as we get use out of it.

PR #120, opened 2021-07-13
Bring back explicit `map` and `list` keywords for type defns

rvagg:
The shorthand is annoying @warpfork enough that he wants these back (they were originally). It is easier to parse this way, but it makes them slightly more verbose. The most notable verbosity in the changes below is `type MyMap map { String : &Any } representation advanced ShardedMap`, whoa boy.

This doesn't do away with the inline definition style, just when you're declaring a `type`.

There's also a matter of `type FooLink &Foo` which is now an outlier .. should it be `type FooLink link &Foo` or something more verbose like `type FooLink link { expectedType Foo }`, which would more explicitly match schema-schema, but would also drive you to use inline forms because this might be unreasonably verbose.

### DIFF
--- a/_legacy/specs/block-layer/graphsync/known_extensions.md
+++ b/_legacy/specs/block-layer/graphsync/known_extensions.md
@@ -18,7 +18,7 @@ When a requestor sends a request, it should send a CBOR encoded IPLD Node that i
 The IPLD schema is as follows:
 
 ```ipldsch
-type DoNotSendCids [Cid]
+type DoNotSendCids list [Cid]
 ```
 
 The responder node will execute the selector query as it would normally. However, if it supports the extension, when the selector query passes over any blocks that have a cid from the DoNotSend list, the responder will not send that block back, knowing ahead of time the requestor already has it. The responder does not send a value back in the response for this extension.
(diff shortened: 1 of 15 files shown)


warpfork:
Yeah, I still like this.  It makes my brain's parser go less kerthunk.

I can imagine another brain going kerthunk when it reads an inline map declaration for the first time, and there's no keyword.  But I've also seen a brain going kerthunk the first time they read `type Foo {X:Y}` until it's explained that `{` is a sigil here.  I'd prefer to optimize for the latter of the two, because it's an earlier impression.

Somehow I don't mind `type FooLink &Foo` as much.  I can't quite put my finger on why.  Perhaps because `&` doesn't show up anywhere else in the DSL, but `{` does (for grouping, and thus shows up _often_), and that makes `{` feel weirder.  `type FooLink link &Foo` doesn't seem offensive either, but leaving it as it is doesn't bother me.

warpfork (approved):

rvagg:
So, to be honest, now I see this in PR form, I don't love it. I might let this sit here until some work can be done on either the JS or Go parsers to see if it's compellingly better for that

rvagg:
Closing this for now, we're moving too far into schema usage to change this easily now - although it could be brought back in as a optional tokens in these positions. Currently nobody is keen enough to push on this.

PR #121, opened 2021-07-13
Add stringprefix union representation strategy

rvagg:
From @warpfork's wishlist


rvagg:
rebased

@warpfork you said you were working on this too? do you have adjustments, alternative language or a replacement to this? or feel free to just push into this branch if you want as well. This is your representation strategy after all, I was just filling in gaps.

warpfork:
I did race you this back in July, yes :sob:  My fail.  I think you produced much better docs than what ever happened in what I was slow to push, so I'm merging yours :)

PR #128, opened 2021-08-11
Add dag-pb investigation doc

rvagg:
Readable version @ https://github.com/ipld/ipld/blob/rvagg/dag-pb-investigation/design/tricky-choices/dag-pb-forms-impl-and-use.md

This relates to https://github.com/ipfs/go-ipfs/pull/7976 and is intended to help with some decision making, or at a minimum, to help build a brief summary of changes for users.

There's a lot in here but for those who are deep in this level of the stack I think it'll be informative, and we have some choices to make coming out of this.

@warpfork @aschmahmann @Stebalien @achingbrain I think you might want to all cast an eye over this. Sorry that it's so long, but there's a lot of history and implementation detail being covered! There's a couple of options outlined at the bottom for moving forward, but I'm sure there are other options we could come up with.

### DIFF
--- a/design/tricky-choices/dag-pb-forms-impl-and-use.md
+++ b/design/tricky-choices/dag-pb-forms-impl-and-use.md
@@ -1,0 +1,126 @@
+# An Investigation Into DAG-PB Implementations and Use
+
+***August 2020 by @rvagg***
+
+This document is a result of some digital archaeology initially spurred by the differences introduced while porting go-ipfs' `dag` commands to use go-ipld-prime, and therefore the latest dag-pb codec, causing differences in putput: `Data`/`Links` (new) vs `data`/`links` (old). It turns out that there's a deep rabbit hole here.
+
+We document [some of this in the dag-pb spec](https://ipld.io/specs/codecs/dag-pb/spec/#alternative-legacy-pathing) but only as it relates to pathing. The fact that we’ve mostly siloed into our separate language stacks meant that the gulf between JS & Go and how we view these things hasn’t been clear. Until now, since we’re integrating go-ipld-prime into go-ipfs and it’s bringing in the new unified dag-pb codec work, where [js-dag-pb](https://github.com/ipld/js-dag-pb) and [go-codec-dagpb](https://github.com/ipld/go-codec-dagpb) are fundamentally doing the same thing under the hood, as per much clarified dag-pb specification, and we’re finally exposing a more “pure” data model up through the `dag` APIs in go-ipfs.
+
+## In the begining
+
+dag-pb (I don't know when it started being called that but it was originally just the data serialisation form) was originally defined as a protobuf schema, which you can find [here](https://ipld.io/specs/codecs/dag-pb/spec/#serial-format). With `Data`  and `Links` on the `PBNode`, and the `PBLink`s have `Tsize`, `Name`, `Hash`.
+
+## Go data forms
+
+go-ipfs defers all its dag-pb concerns to [go-merkledag](https://github.com/ipfs/go-merkledag/). Which in turn serves as a translation layer between the serialized forms and the in-memory forms of these nodes. Serialization happens in [go-merkledag/pb](https://github.com/ipfs/go-merkledag/tree/master/pb) which started life as a protobuf codegen but has been manually tweaked since then my a few of us for various reasons.
+
+go-merkledag/pb has `PBNode` and `PBLink`, matching the protobuf schema.
+
+go-merkledag, on the other hand, has `ProtoNode`, but uses [go-ipld-format](https://github.com/ipfs/go-ipld-format) for its `ipld.Link` type. `ProtoNode` uses lower-case `data` and `links` and does explicit translation when converting to and from `PBNode`. It also has JSON marshal and unmarshallers that are careful to retain these names as they are (if you're not a Go person, keep in mind that Go uses the capitalisation of the first character of variable names as an indicator of public accessibility of struct fields, so there's often hoops you have to jump through to make things appear publically as lower-case, including JSON serialization for which you have to write custom code to access those lower-case fields, or do other funky stuff with tags).
+
+_But_, it defers to go-ipld-format for `ipld.Link` which does some translation of its own. It keeps the upper-case first character and `Name` is the same, but `Tsize` becomes `Size` and `Hash` becomes `Cid`. There's no special JSON handling because these properties are publically accessible (aside from `Cid` custom JSON marshalling from go-cid), but two of them are different from the PBLink form already.
+
+go-ipfs has been relying on JSONification for its `dag get` API for the most part. If you `dag get CID` then it's going to JSON print the object it gets back, so it's up to the codec to get its self in order for JSON marshalling. If you do that for a dag-pb block then you get `{"data":"base64bytes","links":[...]}` because `ProtoNode` sets itself up for that. The `"links"` array is going to have entries of `ipld.Link` forms, like so: `[{"Name":"name","Size":x,"Cid":{"/":"Qm..."}},...]`. It will also work in the reverse, so you can feed it that form of JSON and it'll unmarshal as a `ProtoNode` and its `ipld.Link`s as required.
+
+Interestingly (for me at least), the `"data"` fields are base64, because that's what the Go JSON encoder chooses to do with bytes, perhaps uniquely. From [encoding/json#Marshal](https://pkg.go.dev/encoding/json#Marshal):
+
+> []byte encodes as a base64-encoded string
+
+This is "standard" base64, which comes _with_ padding. We've created [dag-json](https://ipld.io/specs/codecs/dag-json/) which has only really been properly solidified this year (even though the basics have been in place for a while) that formalises bytes in a similar way, but uses _unpadded_ base64 and nests them under a structure to reduce potential namespace collisions (`{"/":{"bytes":"base64here"}}`).
+
+## JS data forms
+
+The JS IPLD stack has arrived at using pure JavaScript objects to represent the IPLD data model. So for dag-pb, you get an object that can be JSONified and also created from a JSON parse. We don't have funky `[]byte` handling in JavaScript which is (I believe) one of the forces that created dag-json, so our dag-json encoder and decoder will do its special byte handling.
+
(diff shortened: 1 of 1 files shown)


willscott:
I'm going to be a proponent of 'option 1'

the filecoin risk and the IPFS work proposed in option 2 seem not worth it to me

design/tricky-choices/dag-pb-forms-impl-and-use.md
```diff
@@ -0,0 +1,126 @@
...
+
+## In the begining
+
+dag-pb (I don't know when it started being called that but it was originally just the data serialisation form) was originally defined as a protobuf schema, which you can find [here](https://ipld.io/specs/codecs/dag-pb/spec/#serial-format). With `Data`  and `Links` on the `PBNode`, and the `PBLink`s have `Tsize`, `Name`, `Hash`.
+
+## Go data forms
+
+go-ipfs defers all its dag-pb concerns to [go-merkledag](https://github.com/ipfs/go-merkledag/). Which in turn serves as a translation layer between the serialized forms and the in-memory forms of these nodes. Serialization happens in [go-merkledag/pb](https://github.com/ipfs/go-merkledag/tree/master/pb) which started life as a protobuf codegen but has been manually tweaked since then my a few of us for various reasons.
```
olizilla, line 15:
```suggestion
go-ipfs defers all its dag-pb concerns to [go-merkledag](https://github.com/ipfs/go-merkledag/). Which in turn serves as a translation layer between the serialized forms and the in-memory forms of these nodes. Serialization happens in [go-merkledag/pb](https://github.com/ipfs/go-merkledag/tree/master/pb) which started life as a protobuf codegen but has been manually tweaked since then by a few of us for various reasons.
```

design/tricky-choices/dag-pb-forms-impl-and-use.md
```diff
@@ -0,0 +1,126 @@
...
+* Resolving all of the data model pieces of a dag-pb node is possible with the `/ipld/` prefix (i.e. even though you can currently print a JSON version of a `ProtoNode` and its `ipld.Link`s, you can't do a whole lot with them without passing them on to some other tool).
+* Custom JSON printing is gone (it's gone anyway with the go-ipld-prime refactor for `dag get`) and we have pure dag-json printing now
+* JavaScript's ability to `dag get Qm.../Links/0/Hash/foo` would need to go and support named links only for bare pathing.
+  * js-ipfs needs to implement named link resolution somewhere new, since it's now not available at the codec level (perhaps something similar to the go-unixfsnode layering ...) and we don't have any `resolve()` in our codecs
+* js-ipfs needs `/ipld/` functionality for data model pathing
+* js-ipfs needs to stop forcing dag-pb results from `dag get` back through the dag-pb codec, just output what is resolved in the codec requested
+* js-ipfs needs other new goodness introduced in go-ipld-prime in go-ipfs
+
+### Option 2: We go the data model form wrong
```
olizilla, line 95:
```suggestion
### Option 2: We got the data model form wrong
```

design/tricky-choices/dag-pb-forms-impl-and-use.md
```diff
@@ -0,0 +1,126 @@
...
+
+In https://github.com/ipld/specs/issues/245 and then https://github.com/ipld/specs/pull/247 an IPLD Schema form of dag-pb was merged. There was some discussion about casing of names at https://github.com/ipld/specs/pull/247#discussion_r390220519 where it was resolved that we should match the protobuf schema. We were apparently oblivious to the additional `ProtoNode` representation, _or_ we didn't consider it relevant to the "data model" form of dag-pb. (There is probably some valid critique of the previous siloing between IPFS and IPLD here too since the IPLD team haven't noticed any of these problems until now.)
+
+The IPLD Schema has since been used to rewrite the new JS and Go codecs for dag-pb, but this was not a major departure since these codecs (originally js-ipld-dag-pb and go-merkledag/pb) at their protobuf level were always using these forms. We were adding additiona lenses onto these codecs before they presented themselves at the IPFS `dag` API - via `ProtoNode`, `ipld.Link` and the custom printing forms in js-ipfs `dag get`.
```
olizilla, line 47:
```suggestion
The IPLD Schema has since been used to rewrite the new JS and Go codecs for dag-pb, but this was not a major departure since these codecs (originally js-ipld-dag-pb and go-merkledag/pb) at their protobuf level were always using these forms. We were adding additional lenses onto these codecs before they presented themselves at the IPFS `dag` API - via `ProtoNode`, `ipld.Link` and the custom printing forms in js-ipfs `dag get`.
```

design/tricky-choices/dag-pb-forms-impl-and-use.md
```diff
@@ -0,0 +1,126 @@
...
+3. go-ipld-prime handles path resolution / traversals for `dag get`
+4. There is special-casing, via go-unixfsnode, such that loading a dag-pb node will attempt to reify it into a [UnixFS data model form which has a custom `LookupByString`](https://github.com/ipfs/go-unixfsnode/blob/main/pathpbnode.go) that will search its dag-pb node `Links` array for a named link and return that CID if it has one by that name.
+5. The custom `LookupByString` means that you can't resolve anything else in the data model form of the dag-pb node.
+6. There is an `/ipld/` override for resolution, such that when the path is prefixed with this, will _not_ perform the unixfsnode reification and you'll get the "pure" data model form and can resolve all of the properties of `{Data,Links:[{Name,Tsize,Hash},...]}` in a standard pathing form.
+
+Worth noting is that the current (pre go-ipld-prime) version of go-ipfs **does not do anything special with the `/ipld/` path prefix and dag-pb nodes**. It appears to behave the same as without, resolving named links and nothing else.
```
aschmahmann, line 70: This is close, but I don't think quite accurate. IIUC what's happening both pre and post go-ipld-prime with respect to /ipfs and /ipld is the following:

Pre go-ipld-prime:
- /ipfs -> Interpret the data as UnixFS, meaning QmFoo/NamedItemInAHAMTDirectory/bar` works
- /ipld -> Interpret the data as non-IPLD, but with ["legacy pathing"](https://github.com/ipld/specs/blob/a652a23ba974c7b524882602a12b6d978b13829a/block-layer/codecs/dag-pb.md#alternative-legacy-pathing) that traverses the named links and hides the size and data fields. In go-ipld-prime world this is analogous to using [PathPBNode](https://github.com/ipfs/go-unixfsnode/blob/d6733f6db69431075c9cf7b30e0ee1f6a082f4b0/pathpbnode.go) as a lens.

Post go-ipld-prime (branch today):
- /ipfs -> Interpret the data as UnixFS
- /ipld -> Interpret the data as plain as possible without any legacy pathing

Main takeaways from an investigative view (i.e. not judging the outcomes):
- /ipld did used to do something in go-ipfs, go-ipld-prime is changing the behavior
- Keeping /ipfs is not a reliable way for us to do named pathing through DagPB objects since we'll use the UnixFS children instead of the object links (e.g. no accessing of named HAMT shard links). Additionally, if we add more strictness to the UnixFS reifier then even basic non-UnixFS DagPB nodes might be untraversable this way.
- As it stands with go-ipld-prime we have killed off named path traversal of DagPB objects via `ipfs dag get`

olizilla (commented):
Good digging! i mearly offer some typo fixes.

> the `data` property is going to end up as an ugly associative array: `{"data":{"0":115,"1":111,"2":109,"3":101,"4":32,"5":100,"6":97,"7":116,"8":97},"links":[...]}`

is a familiar sight from hacking on explore.ipld.io

BigLep:
Thanks for the write up.  I read through it, but I know I'm not catching all the implications.  My hope is that we end in a world that is simpler/more intuitive.  It sounds like there is followup work here depending on the option.  Please make sure the corresponding issues get created/linked.

design/tricky-choices/dag-pb-forms-impl-and-use.md
rvagg: I don't quite understand the last two points you have here @aschmahmann.

From what I understand, and from experimenting, it looks like `QmFoo/NamedLink/foo` _and_ `QmFoo/NamedShardedLink/foo` works old and new, it's just that the output is now dag-json and has the TitleCased `Data` and `Links`. Aside from that, the difference is in `/ipld`, as you've pointed out, works for named links with the old, but not sharded named links. I can do `/ipld/QmFoo/NamedLink` but not `/ipld/QmFoo/NamedShardedLink` but I can do `/ipld/QmFoo/AA`, `AA` being one of the names that the HAMT is going to give for its sharded named links.

So with the go-ipld-prime changes, the only loss (aside from json changes) is the ability to use unsharded named links on sharded directories (`AA` etc.). But you still have the ability to access those by index, so `/ipld/QmFoo/Links/170` is going to get me the same as `/ipld/QmFoo/AA` would have got me before.

Which is all pretty obscure and doesn't seem like a loss of functionality for the vast majority of users, likely no users since this is all hidden (including behind a flag in go-ipfs!).

I'm going to update the doc to talk about sharded directories because that's a difference I didn't investigate yesterday.

rvagg:
Added some additional notes about sharding and `/ipld/` semantics in the latest commit. Also added a slightly rambly section at the end with "Option 3" that's a kind of compromise but IMO is worst of both worlds. I wanted to introduce it here because ADLs, which we are already using via go-unixfsnode, provide us with a way to morph the data model into new forms which we could do to go from `Data`, `Links` and friends to `data`, `links` and friends. But .. I think we'd just end up creating more problems for ourselves and confusing users even more as they look behind each successive curtain.

aschmahmann:
> "Option 3" that's a kind of compromise but IMO is worst of both worlds

Agreed. Option 1 seems like the most sane thing to do. It continues to give us UnixFS support via `/ipfs` and non-magical and understandable pathing support via `/ipld`. The edge cases around named DagPB links are a bit of a shame, but they were pretty confusing anyway and AFAICT not particularly used.

warpfork:
A lot of digital ink has already been spilled on this, so I'll keep my comment on choices here to a short "2 cents":

- I think the TitleCase is unfortunate, and weird.  It's a golang naming convention that escaped to affect our datamodel/json, and if we were in a greenfield, I'd say that's bad.
- That said, I frankly don't have a lot of spoons for caring about dag-pb.  And we're not in a greenfield.  So it can be weird all it wants.  And I'm on board with work-minimization.
- The rest of the remarks in Option 1 about moving towards consistency in datamodel pathing vs named links pathing checks out to me.  :white_check_mark:

PR #130, opened 2021-08-18
Renovate Selector specs to use testmark format.

warpfork:
The fixture data is now contained in markdown code blocks.

This is both machine parsable, and at the same time, should render nicely (and even with syntax highlighting!), if you're looking at this in the website.

I believe this is going to markedly increase the _quality_ of our docs (we can put prose descriptions right next to the hard data!  and there's no split of wasted effort between docs and specs!), increase the _number of eyes_ that *reach* docs and examples written this way (people don't read ugly fixture files willingly!), increase the _confidence_ we have in the docs (the readable stuff will be load-bearing and directly tested!), and increase the _joy_ and sense of reward we get from writing fixtures (they can be pretty, and you can hand people links to them, even to specific rendered sections!).  All big wins.

I've made a library for parsing this, and designated it the [testmark format](https://github.com/warpfork/go-testmark).  While I've only written this in Go so far, I think it will be easy to replicate in other languages, and it should be uncontentious to do so, because we're really just piggybacking on established markdown syntaxes here.

Removed the previous ".taf" file.  (I wrote the taffy format.  I'll slay it, decisively, too.  Taffy was a dumb idea.  Renderability matters.  This markdown-based format, "testmark", will serve this better.)


warpfork:
I'm now using these in https://github.com/ipld/go-ipld-prime/pull/231, and it works like a _charm_.

Merging.

rvagg:
this is not showing up on ipld.io, I wanted to see what it looks like there but it's got the old stuff, @warpfork can you debug?

warpfork:
Huh.  You're right.  Kicking fleek in the shins, seems something hung during the deploy process.

... There we go: https://ipld.io/specs/selectors/fixtures/selector-fixtures-1/

PR #134, opened 2021-09-06
add cross-codec testmark fixtures

rvagg:
Per https://github.com/ipld/codec-fixtures/pull/13

So, tbqh, this is an attempt to make codec-fixtures more palatable to @warpfork who wants everything in this repo and would prefer fixtures in testmark format, but it's also kind of neat.

These files are generated from codec-fixture data as it exists today and is intended to be used according to the same pattern there. An implementation that wants to test one or more codec would load data from these files, parse data using the codec in question into the data model, then spit it out using an alternative codec and make sure the CID matches for that codec. Going both ways increases the coverage and we get to test the ability of a codec to properly parse the data and properly represent it in the data model such that any other codec that supports that form can reencode it and get to the same CID. That's the basic idea of codec-fixtures and now the data is in testmark format here too.

https://github.com/ipld/codec-fixtures/pull/13 also uses testmark's `patch` feature to update the doc and can ammend new fixtures to the bottom of the doc. Intermediate text can be manually edited where notes are helpful. I've put an identical explanation at the top of each of the files included here.

Each fixture has the bytes for the codec in question in hex, plus the CIDs for those bytes in all other supported codecs for that data. dag-json is a special case in that it also has the string form, so you can see what's in the fixture, which is nice. dag-pb and dag-jose have a subset of the data since they can only represent a limited subset of the fixture forms, obviously.

It would be possible to flip the source of the data so that these testmark files contain the authoritative version and the test runners (now run by cron) in codec-fixtures use these files rather than having copies. The test runners could also be moved into this repo. I'm just not personally really keen on beefing this repo up with code.

Feedback requested. Good idea, or no?


warpfork:
This looks _super_ appealing to me, and I'm _so_ excited you did this.

All of the things I would consider possibly tweaking are very mild and mostly stylistic.

Relatively structural and important:

- Use the "directory-like" convention for the hunk names to make it easier for us to interact with it in organized ways.
  - I usually make things follow a convention of "`{fixture-name-here}/{topic-name}.{format-hint-if-applicable}", and that's been working out pretty well.
    - Maybe something like `jwe-asymmetric/data.dag-jose` + `jwe-asymmetric/data.dag-jose.cid` + `jwe-asymmetric/data.dag-json.cid`, etc, to riff on ironing out one of the examples I see already in the diff.
    - I don't have strong opinions about the "format hint" vs "topic name" part.  That's just "usually" made sense in other domains I've been using this format.  It doesn't seem to apply super cleanly here, since the whole point of these fixtures is to prance across codecs.  I dunno -- try it on for size?; if it doesn't fit, ignore me on this bit.
  
Stylistically:

- I tend to go a little lighter on the headings.  I usually find it reads well to use approximately one heading (usually, h3) per "directory" of data hunks (which also maps to roughly one "fixture"'s worth of interrelated data).
  - I just use a quick line or two of plain prose to say what things are, if it feels appropriate.
  - ... but this is a nit.  You can ignore it.  Do whatever feels good.
- I think we've got enough data here that breaking it up into a couple files could be reasonable?
  - Plonking a decent amount of stuff in one big file is fine; makes for easy reading and easy editing.
  - Some of these things like the fixtures that are sizable and say "garbage" in the name?  Maybe put those in a separate file, IMO; someone's less likely to _read_ those as a human.

Overall:  I'd love to move forward with this.  :bow:

Re: code: I'd be perfectly happy to see code appear in the go-ipld-prime repo that references this.  (So, in other words, agree with your comment about keeping code out of this repo.  If we have code to move, I'd rather move it closer to my implementation repos anyway.)

rvagg:
I thought you might take issue with my `-` rather than `/`.. I've switched it now so it has the pattern `<name>/<codec>/<what>` where `<what>` is either `bytes` or `cid`. So if you were to compile all of these together you'd get `<name>` fixture with each supported `<codec>` and both its `<bytes>` and `<cid>`.

I've stepped down a heading level and just bolded the sections, although my preferences is usually for embedding the structure using markdown where it exists. In this case it's not really a big deal because in truth these aren't really meant to be casually read! I did imagine future updates adding notes about particular cases, to say what's going on and why it's special.

Re breaking it up - there's not really one obvious way to delineate them, and the source data doesn't so unless I go back and do it there, or flip to making this the authoritative source, then it'll difficult to pull off cleanly. I think I'd prefer to punt that one down the road a little.

I had to do some jiggering with frontmatter to get the template engine to ignore all the curly braces in the dag-json string output blocks, but it's compiling now, current version @ https://bafybeid63ywauramdwt2gngt5n6gnmgnrm4h5slwktdl7hf6twubzutt5q.on.fleek.co/specs/codecs/dag-cbor/fixtures/cross-codec/

Unfortunately the rendered version doesn't do any kind of auto-break on the hex output. I was originally thinking of having a fixed line width and breaking on that, so I might have to go back and do that so these pages don't get out of hand with the larger blocks. Test runners would just have to strip out newlines to reconstitute the hex data.

rvagg:
Added linebreak for hex output @ 120 chars which formats a bit nicer

See https://bafybeihj3udp2stvrm57i3nxpebwx54ot5y63qynkc6rcg4ayj3s4wuuvi.on.fleek.co/specs/codecs/dag-json/fixtures/cross-codec/ for current build

Interestingly, the `json` codeblocks get a proper scrollable frame, while the bare ones don't, something to do with the syntax highlighter I suppose but the treatment is quite different, as seen in the dag-json link above:

Screenshot 2021-09-09 at 2 21 17 pm

Not super pretty, but they are test fixtures after all.

rvagg:
removed dag-jose from here, it's not resolved upstream (i.e. breaking)

PR #136, opened 2021-09-09
Schema schema synchronization

warpfork:
This is... a bulk fix to every inconsistency and bug in the schema-schema I've been able to find today.
On the sad side: hooo boy, is it a diff.
On the bright side: I... think should be every fix we'll need for a good while.  It fixes every bug and nit I've accumulated a note about in the last year or more.

The diff overall may be somewhat intimidating -- especially because in the JSON form, one of the changes causes an indentation level shift -- but if you go through it commit by commit, or focus attention on the DSL (which *didn't* have an indentation change) it should be much less terrifying.

The commit log is the best reference for what happened, but also, in brief, here was my checklist as I ripped through this:

- [x] bugfix: the `Type` declaration in json was missing its "members" field entirely, which was clearly invalid.
	- wait, WHAT?  We didn't actually define unions as having a members field?!???!?
	- so this is actually wrong _everywhere_.
	- okay fix all of it then, including the defn of 'TypeUnion' itself.
- [x] bugfix: the `EnumRepresentation` same
- [x] bugfix: the `TypeKind` enum is now all lowercase, instead of undesirably using TitleCase.
- [x] bugfix: the `RepresentationKind` enum is now all lowercase, instead of undesirably using TitleCase.
- [x] `EnumValue` is renamed to `EnumMember`, which says what it is better.
- [x] `TypeTerm` is renamed.  It's now `TypeNameOrInlineDefn`, which says what it is better.
- [x] fix the lack of ability to describe anonymous typed links in unions, and in other inline defns.
	- grabbed the diffs from https://github.com/ipld/ipld/pull/104 and ran away with them.
		- I solved this with a new type called `UnionMember`, plus `UnionMemberInlineDefn`.
		- n.b. some of the union representations use this new `UnionMember` type; some do not, because they can't actually contain values which would need it in order to be described.  (Inlines only work for members which are structs, and thus must be named; Stringprefix only works for strings, which must be named; etc.  There are additional details of validating those rules that is still pushed to the logical level, but there's no reason to let the schema describe things that are clearly  invalid, either.)
- [x] `unit` is now a typekind.
- [x] `unit` has representation strategies `null` `true` `false` and `emptymap`.
- [x] `any` is now a typekind.
	- amusingly, it's also defined using `unit`, of all things.  (Makes sense: there's nothing to describe about `any`, because that's how much of a wildcard it is.)
	  - [x] revisit.  It could equally well be describe using `struct {}`, as most other simple types do.  It might be hewing better to the "keep schema-schema using the minimal number of features" goal to define it this way.
	    - yep, done.  minimizing surface area is indeed desirable.
- [x] `any` doesn't have any representation strategies because the whole point is that it's a wildcard that basically allows any datamodel primitive and doesn't mess with them.
- [x] stringprefix: it should support a joiner?  Undecided, input wanted.
	- Our Go API had such an explicit joiner string, but we've mostly switched to thinking that was a misstep.  But it could also be fine to support, as long as it remains optional and an empty string is also accepted: I've realized you can get better error messages out of having a standard delimiter.  Not sure where this lies on cost/benefit.
	- May want to shift the whole thing to have a struct that contains an anonymous map just to make sure we can add that later as a smooth schema evolution if we want.
	- Decision: yes on the struct, so we have evolvability space reserved; no on actually putting a string field there at this time.
- [x] check on bytesprefix defn being correct, multibyte, etc, now.  (I think Rod might've already done this.  (Yes, confirmed, he did.  Thank you diligent sir.))
- [x] added `HexString` type as a marker for what's going on within bytesprefix.  (Not a functional change; mostly for clarity of communication.  It's still just a string kind, ultimately.)
- [x] the `Type` union is using keyed representation strategy, instead of inline.
	- Making an executive call here: we're going to keyed unions now. (Sorry @rvagg; I know you're still on the fense about this.  I'm pushing us over.  @mvdan also ran into this as a headscratcher because one of the trickiest things in this whole system to implement is inline unions.  I'm making it an explicit design rule for the schema-schema now: the schema-schema should intentionally use only as many of the features of schemas as necessary, and prefer the ones that we feel are reasonable to expect to be implemented early in a new IPLD library in a new language.  That means inline unions are _out_.)
	- this causes the majority of the diff, because it changes indentation
- [x] the `InlineDefn` and `UnionMemberInlineDefn` types are similarly switch to keyed strategy instead of inline.
- [x] Should we rename all these "kind" fields to "typekind"?  That would be in keeping with a resolution we made about increasing clarity in that terminology in this last year.
	- Nope; Irrelevant!  The switch to keyed unions nukes all of these instances of the string "kind" which were questionable.
- [x] Added description of prelude, and its declarations, to specs directory, in a new file.
	- Technically not causally relevant to most of the rest of this PR, but it's high time to just lay this out.  (And it shows where and why `any` typekind are important!)
- [x] `TypeEnum.members` is a map trying to act like a set.  We just introduced `TypeUnion.members` as a list.  Should we fix TypeEnum to use a list too, to be consistent?  ~Probably.~  Yes.
- [x] Consider: renaming `Type` to `TypeDefn`, which would say what it is better.
  - yes, gonna do this.
  - also, rename all `Type{typekind}` to `TypeDefn{typekind}` (and also `TypeCopy`, that odd duckling, to `TypeDefnCopy`).
  - Justification will be included in the commit message.
- [x] the `SchemaMap` type can be discarded; an anonymous map in the `Schema` struct is sufficient.  (Otherwise: I would've at least felt compelled to rename it `TypeDefnMap`.)
- [x] bugfix: `TypeLink.expectedType` is `TypeName` not `String`.
- [x] a heap of misc docs improvements.
- [x] default representation for maps and lists is communicated by absent 'representation' field.  (Is a schema-schema fix; we'd been doing it this way already.  Whoopsie.)

As I post this, I have a few unchecked checkboxes above.  All of those are things I think I still want to get done before landing this.  

I'd also like to land it... promptly.  A driver for tackling this work today is @mvdan's work on mechanical parsing of schema DMT data in our go libraries, and I want to get this all to solid ground so he can do that work once, rather than once and then immediately have to patch a bunch of it.  (Well, as much as possible.  I may actually be too late for that.  Still!  The iron is relatively hot!  At the very least it's not quenched yet!  Striking!)

### DIFF
--- a/specs/schemas/schema-schema.ipldsch
+++ b/specs/schemas/schema-schema.ipldsch
@@ -105,6 +105,8 @@ type Type union {
 	| TypeUnion "union"
 	| TypeStruct "struct"
 	| TypeEnum "enum"
+	| TypeUnit "unit"
+	| TypeAny "any"
 	| TypeCopy "copy"
 } representation inline {
 	discriminantKey "kind"
@@ -128,6 +130,8 @@ type TypeKind enum {
 	| Union ("union")
 	| Struct ("struct")
 	| Enum ("enum")
+	| Unit ("unit")
+	| Any ("any")
 }
 
 ## RepresentationKind is similar to TypeKind, but includes only those concepts
@@ -137,8 +141,9 @@ type TypeKind enum {
 ## those concepts are introduced in the IPLD Schema system, and when serialized,
 ## all of them must be transformable to one of these representation kinds
 ## (e.g. a "struct" TypeKind will usually be transformed to a "map"
-## RepresentationKind; "enum" TypeKind are always "string" RepresentationKind;
-## and so on.)
+## RepresentationKind; "enum" TypeKind are typically a "string" RepresentationKind;
+## and so on; exactly what RepresentationKind a type defintion will have
+## is determined by its representation strategy).
 ##
 ## RepresentationKind strings are sometimes used to to indicate part of the
 ## definition in the details of Type; for example, they're used describing
@@ -731,6 +736,60 @@ type EnumRepresentation_String {EnumMember:String}
 ## for each EnumMember (with `Foo ("100")`, those are stored here.
 ##
 type EnumRepresentation_Int {EnumMember:Int}
+
+## TypeUnit describes a type which contains no data at all (other than that fact of its existence).
+## (If this seems strange, consider that the cardinality of TypeBool is 2;
+## the cardinality of TypeUnit is simply 1.)
+##
+type TypeUnit struct {
+	representation UnitRepresentation
+}
+
+## UnitRepresentation is an enum for describing how a TypeUnit should be represented in the data model.
+##
+## Unit types are commonly seen represented in several ways.
+## A null token is a common one.
+## A true token is sometimes seen (especially, when people encode "sets" in json:
+## often this will be seen as a map where the values are keys and the map values are 'true').
+## Also, an empty map can be a useful unit value;
+## an empty map accurately communicates a lack of data.
+## (The emptymap strategy can be a particularly interesting choice if you want to
+## have a schema that is evolvable in the future to start using a struct or map
+## in the same place as the unit type currently stands, while having older documents
+## continue to be parsable by the evolved schema.)
+##
+## Unlike many of the other representation information types seen in the schema-schema,
+## this one is just an enum, rather than being a union.
+## That's because there's no possibility of every needing to annotate more customization
+## onto values in the unit type... because there are no possible values in the unit type.
+##
+## Note that there is no discernible logical difference between
+## `type Foo struct {}` and `type Foo unit representation emptymap`;
+## only that the latter can be said to be a more explicit description of intent.
+## Both will result in identical representations, and both have identical cardinality (which is 1).
+##
+type UnitRepresentation enum {
+	| Null ("null")
+	| True ("true")
+	| False ("false")
+	| Emptymap ("emptymap")
+}
+
+## TypeAny describes a type which can contain data of any kind.
+## It's essentially an escape valve; it says "we don't really know what lies beyond here".
+## The type-level data model kind of an "any" type can be anything;
+## it depends on what the inhabitant value is.
+## The representation-level kind can similarly be anything;
+## it will match whatever the type-level kind is.
+##
+## It is not possible to regain useful types on deeper values after using an 'any' type;
+## from then on, the rest of the data is locked in on having exactly that 'any' type,
+## and having no further ability to have separate type-level and representation-level behaviors.
+##
+## 'any' was introduced as a typekind after discovering it is not possible
+## to emulate its behavior by constructing a union; see
+## https://github.com/ipld/specs/issues/318 for some discussion of this.
+type TypeAny unit representation emptymap
 
 ## TypeCopy describes a special "copy" unit that indicates that a type name
 ## should copy the type descriptor of another type. TypeCopy does not redirect a
--- a/specs/schemas/schema-schema.ipldsch
+++ b/specs/schemas/schema-schema.ipldsch
@@ -114,20 +114,20 @@ type Type union {
 ## Notice this enum's members are the same as the set of strings used as
 ## discriminants in the Type union.
 ##
-## TODO: not actually sure we'll need to declare this.  Only usage is
-## in the Type union representation details?
+## This enum is not actually used elsewhere in the schema-schema,
+## but does correspond to the discriminant values used in the Type union.
 type TypeKind enum {
-	| Bool
-	| String
-	| Bytes
-	| Int
-	| Float
-	| Map
-	| List
-	| Link
-	| Union
-	| Struct
-	| Enum
+	| Bool ("bool")
+	| String ("string")
+	| Bytes ("bytes")
+	| Int ("int")
+	| Float ("float")
+	| Map ("map")
+	| List ("list")
+	| Link ("link")
+	| Union ("union")
+	| Struct ("struct")
+	| Enum ("enum")
 }
 
 ## RepresentationKind is similar to TypeKind, but includes only those concepts
@@ -144,14 +144,14 @@ type TypeKind enum {
 ## definition in the details of Type; for example, they're used describing
 ## some of the detailed behaviors of a "kinded"-style union type.
 type RepresentationKind enum {
-	| Bool
-	| String
-	| Bytes
-	| Int
-	| Float
-	| Map
-	| List
-	| Link
+	| Bool ("bool")
+	| String ("string")
+	| Bytes ("bytes")
+	| Int ("int")
+	| Float ("float")
+	| Map ("map")
+	| List ("list")
+	| Link ("link")
 }
 
 ## AnyScalar defines a union of the basic non-complex kinds.
--- a/specs/schemas/schema-schema.ipldsch
+++ b/specs/schemas/schema-schema.ipldsch
@@ -229,26 +229,23 @@ type TypeDefnMap struct {
 	keyType TypeName
 	valueType TypeNameOrInlineDefn
 	valueNullable Bool (implicit "false")
-	representation MapRepresentation
+	representation optional MapRepresentation
 }
 
 ## MapRepresentation describes how a map type should be mapped onto
 ## its IPLD Data Model representation.
 ##
 ## By default, a map type is also represented as a map in the Data Model,
-## but other representation strategies can be configured.
+## but other representation strategies can be configured
+##
+## Note that the `TypeDefnMap.representation` field is optional --
+## the default behavior is demarcated by the lack of any of these values.
 ##
 type MapRepresentation union {
-	| MapRepresentation_Map "map"
 	| MapRepresentation_StringPairs "stringpairs"
 	| MapRepresentation_ListPairs "listpairs"
 	| AdvancedDataLayoutName "advanced"
 } representation keyed
-
-## MapRepresentation_Map describes that a map should be encoded as
-## a map in the Data Model
-##
-type MapRepresentation_Map struct {}
 
 ## MapRepresentation_StringPairs describes that a map should be encoded as a
 ## string of delimited "k/v" entries, e.g. "k1=v1,k2=v2".
@@ -289,22 +286,19 @@ type MapRepresentation_ListPairs struct {}
 type TypeDefnList struct {
 	valueType TypeNameOrInlineDefn
 	valueNullable Bool (implicit "false")
-	representation ListRepresentation
+	representation optional ListRepresentation
 }
 
 ## ListRepresentation describes how a map type should be mapped onto
 ## its IPLD Data Model representation.  By default a list is a list in the
 ## data model but it may be replaced with an ADL.
 ##
+## Note that the `TypeDefnList.representation` field is optional --
+## the default behavior is demarcated by the lack of any of these values.
+##
 type ListRepresentation union {
-	| ListRepresentation_List "list"
 	| AdvancedDataLayoutName "advanced"
 } representation keyed
-
-## ListRepresentation_List is the default representation for TypeDefnList and
-## will be used implicitly if no representation is specified.
-##
-type ListRepresentation_List struct {}
 
 ## TypeDefnLink describes an IPLD link, which is a content-addressable pointer to more data.
 ## (Links in IPLD are implemented by CIDs: they're a hash that identifies another another block of data,
--- a/specs/schemas/schema-schema.ipldsch
+++ b/specs/schemas/schema-schema.ipldsch
@@ -3,6 +3,33 @@
 ## Yes, it's self-describing! :)
 ## -----
 
+## Schema is a the root element of an IPLD Schema document.
+##
+## A complete (if quite short) Schema might look like this:
+##
+## ```
+## {
+##   "types": {
+##     "MyFooType": {
+##       "type": "string"
+##     }
+##   }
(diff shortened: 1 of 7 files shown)


specs/schemas/schema-schema.ipldsch
```diff
@@ -336,14 +341,30 @@ type TypeLink struct {
 ## Unions can be defined as representing in several different ways: see
 ## the documentation on the UnionRepresentation type for details.
 ##
-## The set of types which the union can contain are specified in a map
-## inside the representation field.  (The key type of the map varies per
-## representation strategy, so it's not possible to keep on this type directly.)
-##
 type TypeUnion struct {
+	members [UnionMember]
```
rvagg, line 345: Can you talk a bit more about why you want this in here, since the information is available in `representation` already so this is a double-up? Is it simply convenience? (It is arguably much more convenient than the current layout, but the duplication bothers me a little).

specs/schemas/schema-schema.ipldsch
```diff
@@ -336,14 +341,30 @@ type TypeLink struct {
...
+	| UnionMemberInlineDefn map
+} representation kinded
+
+## UnionMemberInlineDefn is a very similar purpose to InlineDefn,
+## but found specifically within UnionMember.
+## It only allows describing a link type (and not maps nor lists, as InlineDefn does),
+## which is a constraint applied to union membership largely to make sure
+## if there are errors in processing unions, we can make legible messages about it!
+type UnionMemberInlineDefn union {
+	| TypeLink "link"
```
rvagg, line 365: Justification for limiting this to links and not inline maps and lists too?

rvagg: I think, organisationally and from a documentation perspective, inline definitions for maps and lists is probably not a good idea anyway, but is there reasoning beyond that which prevents this from being technically possible? Maybe document further why it's only links because if it's simply _undesireable_ then we could have requests to broaden this from folks who don't see that.

specs/schemas/schema-schema.ipldsch
```diff
@@ -699,19 +723,73 @@ type EnumRepresentation union {
...
+## the cardinality of TypeUnit is simply 1.)
 ##
-type EnumRepresentation_Int {EnumValue:Int}
+type TypeUnit struct {
+	representation UnitRepresentation
+}
+
+## UnionRepresentation is an enum for describing how a TypeUnit should be represented in the data model.
+##
+## Unit types are commonly seen represented in several ways.
```
rvagg, line 750: I like these docs, makes a pretty good summary case

specs/schemas/schema-schema.ipldsch
```diff
@@ -699,19 +723,73 @@ type EnumRepresentation union {
...
+	| False ("false")
+	| Emptymap ("emptymap")
+}
+
+## TypeAny describes a type which can contain data of any kind.
+## It's essentially an escape valve; it says "we don't really know what lies beyond here".
+## The type-level data model kind of an "any" type can be anything;
+## it depends on what the inhabitant value is.
+## The representation-level kind can similarly be anything;
+## it will match whatever the type-level kind is.
```
rvagg, line 783:
```suggestion
## The representation-level kind can similarly be any data valid according to the IPLD data model;
## it will match whatever the type-level kind is.
```

goes without saying in ipld-prime, but elsewhere, like JS, `any` doesn't let you give me a `Date` or `RegExp` object

specs/schemas/schema-schema.ipldsch
rvagg: I guess this matches enums now, so it makes sense ... I've just taken for granted until now that I need to reach into `representation` to find the members and have code in various places doing this!

specs/schemas/schema-schema.ipldsch
warpfork: Weak, actually.  We _could_ drop this type and use `InlineDefn` as one of the members of `UnionMember`.

Mostly, I just find it kind of scary to think about how I'd format error messages if someone wrote a type like this:

```
type Wow union {
	| {Oof:{Wat:nullable Wot}} map
} representation kinded
```

and it doesn't seem nice to read it; and it also seems like making the DSL parsers deal with that would be headachey.

Maybe it's less bad than I think, though.

rvagg:
I think I'm OK with ditching the inline unions, if I had an objection I can't remember what it was other than schema-schema churn is getting really painful because I have a lot of code now built around this so would prefer to not keep on radically changing elements on a whim, but there be good justification for it.

specs/schemas/schema-schema.ipldsch
warpfork: I buried it in the commit message.  :)  (Read: "I will never stop hating github for hiding commit messages, until they someday, I hope, fix their horrible, horrible UI about this.")

> The rule is: the type definition must still be completely clear about
> what will happen at the type level, even if the representation field
> was stripped out or reset to a blank.  For union types: that means
> the members list must be clear, and we cannot rely on the
> representation strategy maps to contain it at some deeper level.
> 
> Yes, this means we must check for logical consistency between the
> representation strategy specification and the members list.
> But we have exactly the same kind of logical matchup work to do
> for e.g. field names in representations for structs, etc,
> so it is not unreasonable nor surprising to have that here for
> union types too.

specs/schemas/schema-schema.ipldsch
rvagg: yeah, I feel the same gross-out at people doing those kinds of things, _but_ we allow them elsewhere so if we're excluding here then state a reason or it'll creep in in a future date by someone else's hand because they notice the inconsistency and don't see a reason for it

reason could just be something like "legibility and coherence breaks down once we start doing this inside unions, more so than in structs and other places `InlineDefn` is used"

specs/schemas/schema-schema.ipldsch
rvagg: sgtm

specs/schemas/schema-schema.ipldsch
warpfork: hrm, does the term 'kind' not make it explicit enough?

Okay, yeah, I guess I buy that this could be clearer.  The sentence right above probably needs a patch too then, though, as it's saying a very similar thing.

rvagg:
sadly this breaks the magic until I can fix it upstream

```
$ make schema-schema
>>> schema-schema >>>
npm run build:schema-schema

> ipld@1.0.0 build:schema-schema
> npm_config_yes=true npx ipld-schema to-json -t specs/schemas/schema-schema.ipldsch > specs/schemas/schema-schema.ipldsch.json

Error parsing specs/schemas/schema-schema.ipldsch: Error parsing schema @ line:792 col:14: Expected "#", "&", "=", "[", "bool", "bytes", "enum", "float", "int", "link", "list", "map", "null", "string", "struct", "union", "{", or [ \t\n\r] but "u" found.
make: *** [Makefile:13: schema-schema] Error 1
```

fails on discovering `unit`. I think everything else might leave the DSL unchanged though, so hopefully it's a simple matter of just shuffling some things around in the DMT the parsers use.

mvdan:
> I have a lot of code now built around this so would prefer to not keep on radically changing elements on a whim, but there be good justification for it.

Please don't bother with updating go-ipld-schema for now, because I'm now doing the work to connect the dots in ipld-prime, and will likely incorporate a modified version of your parser there :)

mvdan:
From the checkbox list, all this sounds good to me at a high level. Updating schemadmt to this new schema-schema should be less than an hour's worth of work, and I think everything else should just keep on working.

And we already have dag-json test fixtures here, so I should be able to verify against those :man_dancing:

warpfork:
Okay, I'm going to bump the use of `unit` back out.  We can do functionally the same thing (in this case) with `struct {}`, and that's one less thing someone would need to implement to bootstrap a processing of the schema-schema, so let's prefer that.

warpfork:
I sort of regret not making all the discriminant tables in all the union representations have the union member name as keys, when reviewing that today.  It would be consistent and easy for them to all be `{UnionMember:*}`.

But that's a fairly old decision now, and I'm not sure it's important enough to be worth changing.

warpfork:
A note, but something I'm not going to act on today: In reviewing all of the schema-schema stuff, I've come back to thinking `"advanced"` probably does not belong as a representation union member for any of the types.  It was an interesting idea, but I don't think it holds.  The markers for application of an ADL should be a different beast, because they can *stack* with representation strategies.  (For example, there's no reason I shouldn't be able to stack, say, an encryption ADL around a union with a representation strategy of bytesprefix -- and a ton of reasons I might want to do that, and _should_ be able to.)

I'm not taking action on this today because we're (still) not about to make it load-bearing today.  But it's something we'll need to address someday.

warpfork:
The bigbig diff to the JSON -- changing the union strategy -- is now pushed.

This means the overall diff legibility in Github for the PR as a whole is now... well, awful.

If reviewing this, even moreso than before, I'd recommend going commit-by-commit.  I'm keeping those diffs individually quite clean (and including a great deal of commentary in the commit messages; it's _such_ a shame github all but hides this).

warpfork:
I have one last semi-sizable change to drop on this stack: some renames.

`Type` is now `TypeDefn`.   All `Type{typekind}` are now `TypeDefn{typekind}`.  And also, `TypeCopy`, that odd duckling, is now `TypeDefnCopy`.

(And @rvagg, you'll be happy to know, at the same time, this tweak made me finally get comfortable with `TypeDefnCopy`.  I think I'm done shaking my fist at it and threatening to remove it.  And here's why... :drum: )

Here are two pieces of prose, the first written with our old terminology, and the second with the new terms:

> "copy" is not present in the TypeKind enum, while it _is_ present in the Type union.  That's not a typo, and kind of accurately indications the subtlety: you can define a type using TypeCopy, but it's not actually a typekind.

... and now read that paragraph again with the new names:

> "copy" is not present in the TypeKind enum, while it _is_ present in the TypeDefn union.  That's not a typo, and kind of accurately indications the subtlety: you can define a type using TypeDefnCopy, but it's not actually a typekind.

... clearer, right?

These renames are a decent diff size in the schema-schema, but actually have _no effect_ on the actual DMT of schemas.  It's just the type names.  So, the `examples.ipldsch.json` file?  Literally doesn't change.

(We may want to update some of our codebases (for example, the DSL parser internals) to use this terminology, though, just for consistency's sake.  (I think it'll make those codebases have slightly better legibility too.  I hope, anyway!))

warpfork:
I *believe* this is now done.  All the checkmarks in the top post are checked, and I've ran out of new checkboxes to add.

mvdan:
I'll be updating schemadmt in ipld-prime to use this new spec and json, to see if that spots anything unexpected.

specs/schemas/schema-schema.ipldsch.json
```diff
@@ -1,585 +1,825 @@
...
+				"representation": {
+					"map": {}
+				}
+			}
+		},
 		"TypeName": {
-			"kind": "string"
+			"string": {}
 		},
 		"SchemaMap": {
```
rvagg, line 26: this needs to go

rvagg:
ooof, this is rough. I've been through all of the new commits and don't disagree with them, but the changes are going to be hard work for all the code I have in place that touches this stuff, and it's going to be difficult justifying the investment in the JS stack unfortunately

mvdan:
Here's a really nice effect from renaming Type to TypeDefn: it no longer conflicts with `var Type typeSlab` in Go APIs :)

warpfork:
I didn't see that _specifically_ coming, but, nice!  (I did think getting a little bit of separation between "TypeDefn" as the thing these documents do, vs "Type" being a thing that is fully compiled and ready to reason about, was gonna be good, holistically.  I guess this confirms it!)

warpfork:
FWIW / to log the conversation:

@mvdan and I just discussed elsewhere if it's worth trying to write the schema-schema without using the "implicit" feature.

We could decide not to.  It's definitely justifiable under the "schema-schema should use minimal features" heuristic.

It would make all schema documents get kinda noticeably larger though, seeing as how false really is something like 100:1 predominating for nullable, and probably also at least 10:1 for optional on struct fields, so, that's... A trade to weigh.

We're probably going to keep using the "implicit" feature in the schema-schema.  It does raise the bar for features needed to bootstrap a parser that can handle the schema-schema, a bit -- but it's too valuable in the impacts it has on brevity of the DMT for every other schema.

mvdan:
After a few minor fixes in the json and schema here, I can confirm that Go's schemadmt is now able to decode the schema-schema dagjson again. So LGTM at an initial practical level.

warpfork:
With that parse working, I think this is now the best shape it's ever been in.  Mergin'.

mvdan:
> ooof, this is rough. I've been through all of the new commits and don't disagree with them, but the changes are going to be hard work for all the code I have in place that touches this stuff, and it's going to be difficult justifying the investment in the JS stack unfortunately

Apologies for going full blast on this stuff. My focus for this week has been to get a proper "parse schema file and compile to a schema.TypeSystem" API in Go. As part of that, Eric noticed a few blockers we needed to fix in the schema-schema first - and since we were already here, he also did a lot of other cleanup work.

Touch wood that this is the last time we do this for a long time, at least at this scale. We even decided for more verbosity that gives us more backwards compatible ways to add more fields in the future, such as using one-field structs.

I'll be doing the Go side, so if you need help with the JS side, I'm happy to try to do my part. But also no pressure because I know it's a significant amount of unexpected work.

PR #137, opened 2021-09-15
schema-schema: fix quoting typos around implicits in the DSL form.

warpfork:
Which kind of value we have there (aka which member of AnyScalar a
parser should yield) should be obvious from the syntax.

So, quotation marks should be present for strings; for bool literals,
there should be no quotations; etc.

(If the member of AnyScalar that a parser finds doesn't actually match
the kind for the type for the field that the implicit clause is
describing, that's a logical consistency check that that should be
noticed during a "compile" phase -- not necessarily something that
the DSL parser should need to worry about.)

### DIFF
--- a/specs/schemas/schema-schema.ipldsch
+++ b/specs/schemas/schema-schema.ipldsch
@@ -228,7 +228,7 @@ type TypeDefnFloat struct {}
 type TypeDefnMap struct {
 	keyType TypeName
 	valueType TypeNameOrInlineDefn
-	valueNullable Bool (implicit "false")
+	valueNullable Bool (implicit false)
 	representation optional MapRepresentation
 }
 
@@ -285,7 +285,7 @@ type MapRepresentation_ListPairs struct {}
 ##
 type TypeDefnList struct {
 	valueType TypeNameOrInlineDefn
-	valueNullable Bool (implicit "false")
+	valueNullable Bool (implicit false)
 	representation optional ListRepresentation
 }
 
@@ -542,8 +542,8 @@ type FieldName string
 ##
 type StructField struct {
 	type TypeNameOrInlineDefn
-	optional Bool (implicit "false")
-	nullable Bool (implicit "false")
+	optional Bool (implicit false)
+	nullable Bool (implicit false)
 }
 
 ## TypeNameOrInlineDefn is a union of either TypeName or an InlineDefn.


mvdan (approved):

mvdan:
Merging as this is/was blocking me.

rvagg:
@warpfork do you recall this discussion last year about this field and the agreement that we quote all values and infer type at compile time? This is a reversal of that position now.

Can we _please_ not make these shifts so quickly and have a bit of deliberation, they're not trivial anymore and we have a lot of code that builds these assumptions in. I'm not saying I disagree with this one but you've not introduced the requirement that we agree on how to parse `AnyScalar` in the DSL in this position - floats, ints, bytes. Previously, with a String, we could hand that problem off to the consumer of the DMT and not make it a DSL problem since the DSL shouldn't need full logic of a codec in itself. Being able to "infer" isn't as straightforward as it sounds and front-loads that problem to the parser where _maybe_ it shouldn't be (maybe it _should_, but a discussion on that would be good).

mvdan:
If there was such a discussion, I couldn't find it on GitHub, and also don't recall it. Perhaps it happened before I joined, or before I got up to speed with what the schema-schema was.

> Can we _please_ not make these shifts so quickly and have a bit of deliberation

You're right that we could at least have waited 24h to let you review. That's my bad for merging quickly.

We essentially have three ways to deal with implicit scalar values in the DMT:

1) We keep the DMT as simple as possible, making implicit values just strings at the syntax Parse level. At the Compile level, which resolves types via a TypeSystem, we would then parse or evaluate the scalar string as a boolean, integer, etc. Implicit values then are strings in the DSL, such as `"true"`, `"foobar"`, and `"123"`.

2) We represent implicit values as scalars in the DMT, and we rely on syntax information for Parse to know what scalar kind the syntax represents. The DSL syntax is then `true`, `"foobar"`, and `123`. Compile should complain if this inferred kind does not match the field type.

3) Something in between options 1 and 2: we always use strings in the DSL syntax, *and* we also coerce those to scalar kinds at Parse time. This only works if the field type is a direct scalar type, such as `someField Bool (implicit "true")`. It does not work with custom types like `type MyBool Bool [...] someField MyBool (implicit "true")`.

The spec used strings in the syntax, and also represented implicit values [as an AnyScalar union](https://github.com/ipld/ipld/blob/e2f4062abcb85c8d120444b177ebb193804f9711/specs/schemas/schema-schema.ipldsch#L646), so it went for option 3. The go-ipld-schema parser also goes for option 3 by [coercing via direct scalar types](https://github.com/ipld/go-ipld-schema/blob/7cf0c504958f9733938c477b6cee93ff3e6d4591/parser/parsing.go#L554-L580).

However, option 3 does not work in practice with the latest schema-schema spec, as it uses implicit values [with fields which are not directly scalar types](https://github.com/ipld/ipld/blob/e2f4062abcb85c8d120444b177ebb193804f9711/specs/schemas/schema-schema.ipldsch#L323).

I hope this illustrates why I was stuck and why I wanted to merge sooner than later :)

As far as the solution goes, Eric here went for option 2. I think either 1 or 2 are reasonable and give us nice consistency; it's just a matter of whether we want the Parse or Compile stages to do more work, and how we want the DSL syntax to look like for implicit values. I personally have a slight preference for keeping implicit values as "just strings" in Parse and letting Compile do all the lifting, but I also understand the point of making the DMT/DSL a bit stricter. So I'm fine with either approach. Updating [the newly integrated parser](https://github.com/ipld/go-ipld-prime/pull/253) to switch the approach would take fifteen minutes :)

warpfork:
Yeah, sorry @rvagg , I do 100% deserve flak for this and you have my appologies.  I do remember discussing this, and (though I don't remember the venue either) I do remember saying something to the effect of "meh, it'll be easier to parse with quotes everywhere maybe".  And yep, that's apparently no longer my position.  (And yes, @mvdan, this does predate when you've started working on this by a good while, so it's my bad for not mentioning it too.)

I think the numbered breakdown of options @mvdan made just above is basically accurate.

When I looked at this again a few days ago... well, our old guess just didn't look good anymore.  Having quote marks everywhere _had_ seemed like it would be the simplest thing for the parser to deal with, and would communicate accurately that fully understanding that value would be defered until fairly late in the process... and yet that's not what was actually going on, since the schema-schema already had the `AnyScalar` type, and the way that was defined stated pretty clearly that the parser had to figure out what kind of value was being described there, so things weren't really that deferred at all.  The mixture of suggestions seemed weird.

Something like what @mvdan called option 2 seems to be what we get if we take `AnyScalar` being defined as a kinded union and run with it consistently.  I know this wasn't what I had suggested for the DSL syntax earlier, but... now that I'm looking at it all afresh?  I pretty much think earlier-me was being a dope and hadn't paid enough attention to this.  :sweat_smile:

warpfork:
After another sleep on it, I have at least recovered the headspace that made me think "quote everything" might be a good idea, so for completeness, I'll note those thoughts down.  (... but then, yes, I think I'll still manage to demolish them on review, too.)

I think I was imagining that we might have a growing number of novel clauses appearing in representation blocks, and had hypothesized that maybe it'd be easier to deal with this if the parse pattern was always `'(' + repeat(keyword + ' "' + value '" ') + ')'`.

(In other words, I think I was probably thinking we'd end up with something similar to golang's struct tags ([these things](https://gist.github.com/sosedoff/b373623a9572cf1a992486d2d87dcd85#file-1_simple-go-L12-L14)).)

This seemed like it might be desirable because it would let us extend things more easily; let people make custom parsers for the values of such clauses; and it would let us have a hands-off policy in the parser.

However, in revisiting today... do those early hypothetical goals still seem valid?  Let's see:

- We just plain _haven't_ had tons of desires for new or novel clauses popping up.  So that consideration, which was hypothetical from the start, may be best to just scratch out entirely.  If we haven't needed it in all this time, it seems increasingly vanishingly unlikely we ever will.
- We also have not wanted custom parsers for any clauses.  So, same story: that goal... It was probably just a hypothetical one, and we can drop it now.
- If we do want to extend schemas and add more clauses of representation details: we'll probably just keep adding more fields (marked either 'optional' or with 'implicit' values, for smooth migration and backwards compatibility, of course!) to the schema-schema.  This seems pretty clear, now.
  - Because we're talking about just adding more fields to the schema-schema, this implies that if we make parse patterns for literals that line up with Data Model kinds (`true`, `false`, `"string"`, etc), those will also keep being reusable if we add more clauses.
- Having a hands-off policy in the parser... yeah, this just turned out to be *very* untrue, as covered in the earlier comments: with the `AnyScalar` type in the schema-schema, we had *already* described not actually having a hands-off policy in the parser.  And that largely seemed appropriate; so I'd expect we'd do more things like this in the future.  It's just very natural to add a new field to the schema-schema and assume that we can parse it from the DSL regardless of if it's an `int` or a `string`.
  - Pushing work off to the compile phase doesn't appear to provide any opportunity to save work at all, either.

So: nope: none of the goals I can remember that originally steered me to proposing quote-always actually seem relevant anymore.

PR #138, opened 2021-09-20
Update CARv2 MultihashIndexSorted and fully-indexed characteristic spec

masih:
Update CARv2 specification to include a format for MultihashIndexSorted
and fully-indexed characteristic bitfield.

Relates to:
- https://github.com/ipld/go-car/pull/239
- https://github.com/ipld/go-car/pull/217

### DIFF
--- a/specs/transport/car/carv2/index.md
+++ b/specs/transport/car/carv2/index.md
@@ -93,7 +93,9 @@ Following the 11 byte pragma, the CARv2 is a fixed-length sequence of 40 bytes, 
 
 The characteristics bitfield contained within the CARv2 header may be used to indicate certain features of the specific CARv2. All bits in the bitfield will be unset (`0`) by default and only set (`1`) where they are being used to signal a characteristic other than the default.
 
-Currently the characteristics bitfield is not used and should have all bits unset (`0`). Future iterations of this specification may introduce characteristic indicators for features such as:
+The first (i.e. left-most bit) value in characteristics bitfield specifies whether the index represents a full catalog of sections that appear in data payload, referred to as `fully-indexed` characteristic. When this characteristic is set (`1`), the index must include a complete catalog of the section CIDs regardless of whether they are identity CIDs or not.
+
+The reminder of characteristics bitfield is not used and should have all bits unset (`0`). Future iterations of this specification may introduce characteristic indicators for features such as:
 
 * DAG walk ordering—none, depth-first, breadth-first, or via [IPLD Selector](../../../selectors/).
 * De-duplication status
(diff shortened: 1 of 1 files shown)


specs/transport/car/carv2/index.md
```diff
@@ -124,7 +126,7 @@ The first byte(s) of a CARv2 index (at the "Index offset" position) contain an u
 
 As the index only contains the hash digest bytes, other details contained within the block's CID and the length of the block's bytes must be derived by inspecting the initial bytes of the block entry within the data payload.
 
-Indexes **should not include identity hash CIDs**. It is assumed that any use of a CARv2 as a blockstore will return identity CID data immediately by extracting it from the CID, therefore there should be no need to provide indexing for such entries.
+Indexes **should not include identity hash CIDs unless `fully-indexed` characteristic is set**. It is assumed that any use of a CARv2 as a blockstore will return identity CID data immediately by extracting it from the CID, therefore there should be no need to provide indexing for such entries. However, when `fully-indexed` characteristic is set, the blockstore should persist blocks with identity CID into the CARv2 data payload.
```
mvdan, line 129:
```suggestion
Indexes **should not include identity hash CIDs unless the `fully-indexed` characteristic is set**. It is assumed that any use of a CARv2 as a blockstore will return identity CID data immediately by extracting it from the CID, therefore there should be no need to provide indexing for such entries. However, when `fully-indexed` characteristic is set, the blockstore should persist blocks with identity CID into the CARv2 data payload.
```

specs/transport/car/carv2/index.md
```diff
@@ -150,6 +152,26 @@ Each bucket, therefore, takes the following form:
...
+
+* MultihashIndexSorted may contain one or more multihash code-grouped buckets of digests.
+* Multihash code-grouped buckets may further contain one or more length-grouped buckets.
+* Buckets are ordered by multihash code then digest length, and concatenated together to form the index.
+* Each bucket is prefixed with:
+  * a "multihash code" encoded as a 64-bit unsigned little-endian integer indicating the common multihash code for the digests in this bucket; followed by
+  * a "width" encoded as a 32-bit unsigned little-endian integer indicating the common byte length of the combination of hash digests and their offsets in this bucket; followed by
+  * a "count" encoded as a 64-bit unsigned little-endian integer which determines the total number of hash digests (and their offsets) bucket.
```
mvdan, line 167: I wonder if you could also deduplicate the definition of width and count and just refer to IndexSorted. It's a balance for sure, but these are entire copy-pasted lines, so I think I'd lean towards referencing the other section.

mvdan (approved):
probably wait for @rvagg to approve as well, since he's better than me at reviewing technical docs :)

specs/transport/car/carv2/index.md
```diff
@@ -150,6 +152,25 @@ Each bucket, therefore, takes the following form:
...
+* Multihash code-grouped buckets may further contain one or more length-grouped buckets.
+* Buckets are ordered by multihash code then digest length, and concatenated together to form the index.
+* Each bucket is prefixed with:
+  * a "multihash code" encoded as a 64-bit unsigned little-endian integer indicating the common multihash code for the digests in this bucket; followed by
+  * length-grouped bucket structure identical to IndexSorted.
+
+Individual index entries are also identical to the IndexSorted entries. Each bucket, therefore, takes the following form:
+
+```
+| multihash-code (int64) | width (int32) | count (int64) | digest1 | digest1 offset (uint64) | digest2 | digest2 offset (uint64) ...
```
rvagg, line 171: If the leading 3 are all unsigned then you probably should add the `u` here. The notes above say at least the first one is unsigned.
At the same time maybe revisit IndexSorted since it has a mix - I think I wrote that originally but did it based on the original carbs impl and they may have been made unsigned for go-car/v2 now?

rvagg (approved):
nice

PR #139, opened 2021-10-19
specs/schemas: regenerate schema-schema.ipldsch.json

mvdan:
(see commit message)

### DIFF
--- a/specs/schemas/schema-schema.ipldsch.json
+++ b/specs/schemas/schema-schema.ipldsch.json
@@ -12,7 +12,8 @@
 						}
 					},
 					"advanced": {
-						"type": "AdvancedDataLayoutMap"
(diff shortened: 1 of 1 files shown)


mvdan:
Also, my ipld-prime PR is basically ready, so I can post it when this is merged.

macotouenaca (approved):

warpfork (approved):
:ship:

rvagg:
@mvdan is it possible to run this with simple bash? Currently we have https://github.com/ipld/ipld/blob/85ea4c8e2478ac2b77ac3f1dec53063446631c5b/package.json#L13 but it uses the now out of date JS parser which won't even parse the DSL now. That field can take arbitrary bash (including multiple steps with `&&` or `;`) if it's easy enough to do with the Go stack now?

PR #142, opened 2021-10-20
specs/schemas: import and start updating Rod's test fixtures

mvdan:
(see commit messages - please do not squash)

### DIFF
--- a/specs/schemas/tests/struct.yml
+++ b/specs/schemas/tests/struct.yml
@@ -8,20 +8,21 @@ expected: |
   {
     "types": {
       "SimpleStruct": {
-        "kind": "struct",
-        "fields": {
-          "foo": {
-            "type": "Int"
+        "struct": {
+          "fields": {
+            "foo": {
+              "type": "Int"
+            },
+            "bar": {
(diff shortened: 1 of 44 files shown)


specs/schemas/tests/struct.yml
```diff
@@ -0,0 +1,96 @@
...
+        "bar": true,
+        "baz": "this is baz yo"
+      }
+    expected: |
+      {
+        "foo": 100,
+        "bar": true,
+        "baz": "this is baz yo"
+      }
+  - actual: # is this OK? string-to-int
```
warpfork, line 43: Oh, gosh.  I know these are probably coming from the earlier files, but the answer is "no" -- I don't think this kind of coercion is desired behavior at all.

rvagg:
fwiw my current code doesn't make use of the "blocks" section now at all either, they're unmaintained and incomplete and can be removed.

I did some very basic beginnings of a testmark version of the DSL->DMT in these, but didn't get very far, the yaml is gross and hard to work with. But getting them in in their current form and iterating from there seems reasonable.

PR #144, opened 2021-10-21
schema fixtures: testmarkify

warpfork:
Porting schema fixtures from yaml to testmark format.  Reviewing with a fine-tooth comb as I go, and interspersing documentation, since the format permits it.

I'm aiming for a documentation level that's useful while terse for a reader that's implementing schema systems, and dry but survivable for a reader that's a user just wanting to see more detailed examples of how to use the DSL (or DMT) and what it matches.

Some of the impetus is covered in https://github.com/ipld/ipld/pull/142#issuecomment-948247749 and other comments in that PR.

There aren't many changes to the logical content, but where there are, the commit messages per commit will describe it.

The new ported content is going into a dir called "fixtures" because that's what the similar dir is called in the selector specs, and in most of the codec specs that have them.  (We're not 100% consistent on this either, in fairness.  For the CAR specs, the dir is called "fixture" -- no "s".  Ah well.  Aiming towards the center of gravity.)

The new [index file](https://github.com/ipld/ipld/pull/144/files#diff-bee479c30c88f3e536241300d0dcf367add5b030bfbdb43ead01765fcf1e0272) for the directory contains a brief explanation of the format and how the data hunks are labelled, so that might be a good place to start reading from.

This is still a work in progress.  I'm removing the old yaml files as I port them, so files remaining with that extension can be seen as the progress bar.


BigLep:
2022-04-05 conversation: putting this and https://github.com/ipld/go-ipld-prime/pull/381 back to Todo, waiting on someone to have the bandwidth to complete all the fixtures.  @rvagg is interested in picking it up if/when he frees up.

PR #146, opened 2021-10-30
Ordering docs

warpfork:


warpfork:
I'm going to merge these fairly promptly because people in the community are actively looking for these docs, and I want to give them the shiny published website experience (and stop sharing stopgap gists! which is where most of this content comes from) as soon as possible. :)  We can iterate on them more later too.

PR #149, opened 2021-11-10
Signal ADLs in selectors

willscott:
Introduce an `ExploreInterpretAs` wrapper in selectors to allow points in a traversal where the current `Node` should be `Reified` using identified, well-known ADLs.

Example:
```
{
	"~": {
               "c": "unixfs",
               ">": { ".": {} }
        }
}
```

### DIFF
--- a/specs/selectors/index.md
+++ b/specs/selectors/index.md
@@ -183,13 +183,12 @@ type ExploreConditional struct {
 }
 
 ## ExploreInterpretAs is a transformation that attempts to 'reify' the current node
-## using an ADL specified by 'as'. Well-known ADLs are identified by their multicodec
-## identifier, and can be found under the tag 'ipld-adl' at
-## https://github.com/multiformats/multicodec/.
+## using an ADL specified by 'as'. ADLs are recognized by agreed-upon strings,
+## similar to libp2p protocols.
 ## The ExploreInterpretAs reification process may introduce a data-dependant amount
 ## of budget on evaluation based on the specific traversal and ADL implementation.
 type ExploreInterpretAs struct {
-	as int (rename "c")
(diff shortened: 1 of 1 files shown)


specs/selectors/index.md
```diff
@@ -181,6 +182,16 @@ type ExploreConditional struct {
 	next Selector (rename ">")
 }
 
+## ExploreInterpretAs is a transformation that attempts to 'reify' the current node
+## using an ADL specified by 'as'. ADLs are recognized by agreed-upon strings,
+## similar to libp2p protocols.
+## The ExploreInterpretAs reification process may introduce a data-dependant amount
+## of budget on evaluation based on the specific traversal and ADL implementation.
+type ExploreInterpretAs struct {
+	as String (rename "c")
```
warpfork, line 191: I guess I'd consider skipping the rename here considering "as" is already just two bytes?  But don't have a strong opinion of it.

warpfork (approved):
This is pretty straightforward so it's got a :+1: from me to go ahead with.

We should really, really capture some of the design discussion that took place in other mediums, though -- perhaps in another page nearby this area in the repo, perhaps as a notebook page, whatever, just as long as its retained somewhere.  Many, many words were already exchanged about whether this is a string or an int, and I have no doubt people are going to want to reread those deliberations in order to understand why this comes out the way it does.

specs/selectors/index.md
willscott: we're renaming end to '$'. i'm just gonna run with this and say maybe we got a 'more-compact-than-json' representation of this set of symbols :shrug:

willscott:
https://github.com/ipld/ipld/pull/150

PR #151, opened 2021-11-19
dag-jose fixtures

warpfork:
Time to bring dag-jose fixtures home.

The fixture data for the dag-jose itself is taken from https://github.com/ipld/codec-fixtures/pull/11/files, though I've encoded them in a slightly different way -- testmark and hexidecimal.

The other two new pieces of fixture data -- the hunk containing data model paths, and the hunk containing prettyprinted dagjson -- I have generated from the go-dag-jose implementation's current behavior.  They need to be verified for correctness!

One of my hopes here with the data model paths hunk that I've introduced is to just make it very very legible what's expected to be in the data model tree that comes out of the codec.  That hunk should be _especially_ easy to diff and be very very easy to discuss.

### DIFF
--- a/specs/codecs/dag-jose/fixtures/index.md
+++ b/specs/codecs/dag-jose/fixtures/index.md
@@ -1,0 +1,117 @@
+DAG-JOSE Fixtures
+=================
+
+About this document
+-------------------
+
+This document contains test fixtures for the [DAG-JOSE](..) codec, and documentation describing them.
+
+These fixtures are executable by parsing them using the [testmark format](https://github.com/warpfork/go-testmark).
+If you're reading this file as markdown, or rendered to the web,
+know that all of the codeblocks here are data that can be easily programmatically extracted and used to drive tests in your implementation.
+
+These fixtures offer several different kinds of data:
+
+- DAG-JOSE serial data -- in hexidemical.  (DAG-JOSE is a binary format, so hexidemical is easier to work with here than the possibly-unprintable binary raw form.)
+	- These should be used as both an decode fixture, and as an encode fixture (the data should round-trip).
+- The list of paths that should be seen in that data, when parsed as [data model](/docs/data-model/).  (This is a good fixture because it's very human-readable, very easy to test, and also provides a form of documentation.)
+- Prettyprinted [DAG-JSON](/docs/codecs/known/dag-json/) data, which is the DAG-JOSE data, parsed, then re-encoded as DAG-JSON, and pretty-printed.  (This gives deep coverage, because it effectively examines every bit of the [data model](/docs/data-model/) view of the data.)
+
+Fixtures
+--------
+
+### JWS
+
+This is a DAG-JOSE object, in hexidemical:
+
+[testmark]:# (jws/data.dag-jose.hex)
+```
+a2677061796c6f616458240171122089556551c3926679cc52c72e182a5619056a4727409ee93a26
+d05ad727ca11f46a7369676e61747572657381a26970726f7465637465644f7b22616c67223a2245
+64445341227d697369676e61747572655840fbff49e4e65c979955b9196023534913373416a11beb
+fdb256c9146903ddb9c450e287be379ca70a5e7bc039b848fb66d4bd5b96dae986941e04e7968d55
+b505
+```
+
+When it is parsed, we should see these paths within the data
+when we walk over it at the [data model](/docs/data-model/) level:
+
+[testmark]:# (jws/paths)
+```text
+TODO
+```
+
+If we re-encoded this data in [DAG-JSON](/docs/codecs/known/dag-json/)
+(and prettyprint it), we should get this result:
+
+[testmark]:# (jws/data.dag-json-pretty)
+```json
+TODO
+```
+
+### JWE symmetric
+
+This is a DAG-JOSE object, in hexidemical:
+
+[testmark]:# (jwe-symmetric/data.dag-jose.hex)
+```
+a46269764c3d2588b80c8ef02a5ebf308b637461675059900c05b961cc30ac41638029d964486970
+726f746563746564581d7b22616c67223a22646972222c22656e63223a224131323847434d227d6a
+636970686572746578745818dd7a8b5b6f0d1cffab6aa5bcbcc7c81cece4a38377211691
+```
+
+When it is parsed, we should see these paths within the data
+when we walk over it at the [data model](/docs/data-model/) level:
+
+[testmark]:# (jwe-symmetric/paths)
+```text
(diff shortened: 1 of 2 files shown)


warpfork:
https://github.com/ceramicnetwork/go-dag-jose/pull/24 contains the code which I used to generate the data model paths and the prettyified dag-json parts of the fixture suite.

warpfork:
@rvagg , I'd appreciate it if you could take a look at this, and @oed, too, when you have a chance.

The main thing I'd like to accomplish here is getting eyeballs on the data model paths thing, because that seems like a good focal point for review.  I found a lot of other fixture work we'd done kinda hard to "eyeball", so this is my attempt to fix that.  I hope the content there looks boring!  If so: great success.

I'd love to have the javascript implementation of dag-jose do the same checks at some point.

(If you're wondering if there's a deeper reason I'm trying to examine this so closely: there is.  I want to make sure that whatever the javascript implementation offers as fields, and what the golang implementation offers as fields, are the same.  This might not matter on day one -- not as much as the codec serial form itself does, anyway -- but it *will* matter as soon as anyone tries to do things with Selectors in JS, for example.  So it's valuable to get it clear!)

specs/codecs/dag-jose/fixtures/index.md
```diff
@@ -0,0 +1,157 @@
...
+fdb256c9146903ddb9c450e287be379ca70a5e7bc039b848fb66d4bd5b96dae986941e04e7968d55
+b505
+```
+
+When it is parsed, we should see these paths within the data
+when we walk over it at the [data model](/docs/data-model/) level:
+
+[testmark]:# (jws/paths)
+```text
+link
```
warpfork, line 41: This line is one I particularly want to talk about.  Is this desired to be here -- aka, is "link" a field we want to have show up in the data model?  Or should it not be here, and is something that can sometimes be created by a helper function, that derives it on the fly?

(The question from https://github.com/ceramicnetwork/go-dag-jose/pull/23/files#r749773563, in other words.)

rvagg:
IMO these fixtures should follow the codec-fixtures pattern for the other codecs and include at least CIDs of the data model form encoded in the other codecs that support it so we can test the data model instantiation (both in and out), that's what those CIDs are for.

warpfork:
So I was thinking about adding CIDs at first, but then I realized I wasn't exactly sure what that was testing and it began to feel a little capricious, so I dropped them.  I still have mixed feelings though, and could be convinced either way, so I'll work through my thinking:

- Con of including CIDs: the choice of hashing function is arbitrary; this is where the feeling of "capricious" came from.  We can certainly just pick one, and that's no technical problem, but it felt odd.  We would have to document in prose what hash function and multihash indicator code we're using.

- Con of including CIDs: it increases the dependencies of the test code and I'm not sure any of them are relevant.  (The hashing function choice already hints at this; in code, in golang, it would also turn into things like using `LinkSystem` and... that's not the end of the world but it just Feels Weird to need to do it, at least in my first impression while coding these tests up last night.)

- Con of including CIDs: it just doesn't seem to be exercising much.  We already have the fulltext of the data as part of the fixture: so, testing the hash of that is just testing the hasher, not our target, the codec.  (N.b. I could see one exception to this: if a codec is sensitive to a trailing linebreak, that's often a touchy bit, and testing the hash is a way to force testing whether or not that trailing byte is included.  However: doesn't apply here, since I used hexidecimal, which removes the opportunity for that ambiguity.)

- Pro of including CIDs: consistency, I suppose (if we assume they're sometimes decisively used, due to the exception in parens in the last bullet point).

- Pro of including CIDs: exercises that implementations got the multicodec indicator code right!

- Pro of including CIDs: didactics, arguably.  It can be good just to see them around.

... Hm.  Okay, on review, the cons are very mild.  And they're also 100% mitigated if someone just chooses not to check that particular fixture.  Perhaps you're right and we should put CIDs back in here.

smrz2001:
Hey @warpfork, I added all the fixtures from `js-dag-jose` to a separate [PR](https://github.com/ipld/ipld/pull/152).

I can make updates per your and @rvagg's discussion above once I can see what you're looking for with the CIDs.

specs/codecs/dag-jose/fixtures/index.md
smrz2001: Since `link` is deterministically derivable from `payload` and never actually encoded, I personally would have preferred it to come from a helper function that could just "enrich" decoded dag-jose data.

However, I couldn't leave it out of the schema and still `datamodel.Copy` an incoming Node (e.g. dag-json decoded) into the dag-jose builder to apply the latter's validations 😭 

(Just a little context to the decision in `go-dag-jose`)

smrz2001: I could do something similar to the other logic I added that I don't quite like - that for "unflattening" JWE/JWS objects, for which I'm going from dag-json -> unflattened map -> json bytes -> reconstructed dag-json before passing that to dag-jose to encode. (BTW, is there an easier way to do that 😅?)

For decode, I could reconstruct the decoded data and add `link` after the decode but that feels super ugly too 🤷🏼‍♂️

rvagg:
> Con of including CIDs: it just doesn't seem to be exercising much

Byte-perfect round-trippishness is exercised by CIDs, we get to test map ordering and all of the tricky edges by doing this. Having the "fulltext of the data" isn't enough to get us there. The point of the CIDs was that they were for the _other_ codecs encoding the data. So that we can observe that the same data can pass through any of the codecs it's compatible with and remain the same. In this way we can assert that the data model form is consistent and stable, since it gets us to any of the output forms (and back again).

specs/codecs/dag-jose/fixtures/index.md
warpfork: > Since `link` is deterministically derivable from `payload` and never actually encoded, I personally would have preferred it to come from a helper function that could just "enrich" decoded dag-jose data or "sanitize" data to be encoded by removing it before the encode.

Right, exactly -- that's what I was asking after.  That's probably what I would've expected, a priori.

That said, I don't have any reason to have a strong opinion one way or the other.  As long as it's agreed that the block form doesn't have it, and the data model form does, then... the specification is specified, and that's my main concern.

warpfork:
@smrz2001 has just added back in CID fixtures for us.

Sounds like there's alignment and confirmation of where the `link` field is expected to be seen (and not seen), and that was the main thing I wanted to double-check at the start of this.

I think this is good to merge by my figuring.

(N.B.: this is a case for merge, not rebase -- the go-dag-jose repo already has some commits which point at commits in this branch, so we should not rewrite their hashes.)

PR #152, opened 2021-11-20
more dag-jose fixtures

smrz2001:
All fixtures from `https://github.com/ceramicnetwork/js-dag-jose/tree/main/test/__fixtures__`


warpfork:
Nice!

I think I see `link` appearing as a path again, so I'll have [the same questions](https://github.com/ipld/ipld/pull/151#discussion_r752824486) about that as in the parent branch -- but we can handle those there.  So I'll add this in right away.

smrz2001:
Tyty!

PR #155, opened 2021-11-25
improving the ADL signalling page

warpfork:
~(Still very much WIP.)~  Almost not WIP at all yet, but still a few blanks to fill in.

The page talking about different strategies for ADL signalling still needed some love.

This diff enumerates some more strategies, each under their own h3, and should introduce and compare them a bit better than before.

One of the key concepts is also that none of these is "the" strategy; hopefully that comes across more clearly by the time we're done here.

### DIFF
--- a/docs/advanced-data-layouts/signalling.md
+++ b/docs/advanced-data-layouts/signalling.md
@@ -9,15 +9,172 @@ eleventyNavigation:
 Advanced Data Layout Signalling
 ===============================
 
-:::todo
-- This page needs to be substantially rewritten to get to the point better.
-:::
+[Advanced Data Layouts](..) are a feature of IPLD which gives us "lenses" to view data in another way.
+
+This page is about answering the question: "How do I know when to use an ADL to read data?"
 
 ### How do I know when to use an ADL to read data?
 
-We call this "the signalling problem".
+Trick question.  You don't.
 
-In short: you don't.
+Using an ADL to read data is **_always_ optional**.
+If you don't want to use an ADL to read data, it's still legible --
+just as the [plain data model](/docs/data-model/) content,
+as a [codec](/glossary#codec) already gave it to you.
+
+You can also use different ADLs to read the same data.
+It's like putting on a different pair of glasses -- maybe some do
+a better job of making the data legible than others!
+
+### That wasn't helpful!
+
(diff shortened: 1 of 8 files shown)


docs/advanced-data-layouts/signalling.md
```diff
@@ -9,31 +9,44 @@ eleventyNavigation:
...
-then it follows that there's absolutely no way for this data to unambiguously indicate that it needs an ADL in order to be understood.
-If there was, it would imply that there's some kind of "reserved words" in the Data Model,
-which would violate some of our other central goals in IPLD, because it would mean some perfectly normal maps and lists would be invalid IPLD or gain magical meaning that they shouldn't;
-we don't want any of that.
+You can also use different ADLs to read the same data.
+It's like putting on a different pair of glasses -- maybe some do
+a better job of making the data legible than others!
 
-So!  Signalling must come from somewhere else.
+### That wasn't helpful!
```
willscott, line 29: Are these the right first 2 paragraphs of the document for someone looking for 'tell me how to do this?'

would it not be faster to answer their question by leading with the signalling methods below?

smrz2001:
Have just started reading through the updates but they're already soooo helpful! Thanks for taking the time to write all this down!

warpfork:
Couple bits of other feedback from @smrz2001 in other channels, which is probably worth integrating:

- Assuming that identification/naming and signalling do get separate pages -- let's indeed move the former first.  "IMO, identification/naming seems like a higher concept to discuss early, while signaling is a much deeper, more elaborate concept."

Realized that I keep coming to identification/naming second because when one looks at writing a wire format, one usually has to figure out _where_ to put it, slightly before figuring out what words to put inside the signalling part of the protocol.  In other words, one discovers the signalling question even before the identification question.  But that's not really the logical order of topics that you'd want to read them in, necessarily; more the implementation detail discovery order.

warpfork:
I'm going to go ahead merge this.  I've done a little wrap-up work, and placed visible TODO annotations in the areas that still need considerably more content.  While it's not perfect and still doesn't cover _everything_ in the depth that I'd like to, it's _considerably_ better than the state of affairs before these diffs, and there's no need to let website readers suffer through the old less-complete content any longer. :)

PR #156, opened 2021-11-25
fix: add dag-jose fixtures CIDs

smrz2001:


warpfork:
Hoot!

One request: can we put the hash function used in the prose, too?  Otherwise that parameter isn't obvious.

Something like:

> This is the base64url-encoded CID for a DAG-JOSE object, when using SHA-FOOBAR (multihash code 0x111):

warpfork:
... I approve, obviously :laughing:

PR #158, opened 2021-12-09
specs/schemas: update struct-tuple

mvdan:
(see commit message)


mvdan:
For an upcoming PR to ipld-prime adding support for tuple reprs in schema/dsl and schema/dmt.

rvagg (approved):

PR #169, opened 2022-01-07
specs/schemas: add first "any" test

mvdan:
(see commit message)


specs/schemas/tests/any.yml
```diff
@@ -0,0 +1,19 @@
...
+      "SimpleAny": {
+        "any": {}
+      }
+    }
+  }
+blocks:
+  - actual: |
+      100
+    expected: |
+      10
```
warpfork, line 15:
```suggestion
      100
```

xD

rvagg:
Does `&Any` appear in our tests yet? That's going to be a common link type for some applications that we should be sure to support if we don't already.

rvagg (approved):
lgtm sans Eric's fix, although I'd also be fine with you just removing the whole `blocks:` section

mvdan:
> Does `&Any` appear in our tests yet? That's going to be a common link type for some applications that we should be sure to support if we don't already.

copy.yml and link.yml do use them, but they're not enabled and tested just yet.

mvdan:
> lgtm sans Eric's fix, although I'd also be fine with you just removing the whole `blocks:` section

I reckon they may still be useful. I just haven't started using them in the ipld-prime test suite.

PR #170, opened 2022-01-10
selectors: touch up InterpretAs

warpfork:
- The prefix "Explore" on its name feels unnecessary; drop.
  (Doesn't change the protocol structurally.)

- The 'as' field is no longer renamed to 'c' in the serial form.
  (DOES change the protocol structurally; but there's only one
  implementation of this, and we changed it too.
  See https://github.com/ipld/go-ipld-prime/pull/301#discussion_r781266224.)

- Added a bunch of docs.

### DIFF
--- a/specs/selectors/index.md
+++ b/specs/selectors/index.md
@@ -65,7 +65,7 @@ type Selector union {
 	| ExploreUnion "|"
 	| ExploreConditional "&"
 	| ExploreRecursiveEdge "@" # sentinel value; only valid in some positions.
-	| ExploreInterpretAs "~"
+	| InterpretAs "~"
 } representation keyed
 
 ## ExploreAll is similar to a `*` -- it traverses all elements of an array,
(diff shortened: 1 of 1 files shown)


willscott (approved):

rvagg (approved):

specs/selectors/index.md
```diff
@@ -182,13 +182,21 @@ type ExploreConditional struct {
...
 
-## ExploreInterpretAs is a transformation that attempts to 'reify' the current node
-## using an ADL specified by 'as'. ADLs are recognized by agreed-upon strings,
-## similar to libp2p protocols.
-## The ExploreInterpretAs reification process may introduce a data-dependant amount
-## of budget on evaluation based on the specific traversal and ADL implementation.
-type ExploreInterpretAs struct {
-	as String (rename "c")
+## InterpretAs is a clause that instructs the traversal to attempt to 'reify' the current node
+## using an ADL, which is specified by the 'as' field.
```
rvagg, line 186: One thing missing from here is what it means for an ADL to be specified, what is this magic string? Not a critical change now, but we may want to get into that kind of detail at some point in our docs rather than hand-waving it away as a turtles-all-the-way-down concerns.

PR #173, opened 2022-01-26
docs: dag-cbor: expand remarks on density and add remarks on complexity

warpfork:
Expand the brief notes in the docs that overview DAG-CBOR to have more remarks on density, and add remarks on complexity.

These are thanks to a series of prompts and comments from @ianopolous.


vmx (approved):

warpfork:
Thank you for the reviews! :heart:

PR #174, opened 2022-01-26
Update JS implementation of CAR to js-car

listenaddress:
Docs are currently pointed at the deprecated `rvagg/js-datastore-car` instead of the new `js-car`.

Example I added is taken from the `js-car` docs—there may be a tighter/better example that'd make more sense here, not sure?


rvagg:
whoops! that's _old_. Thanks for catching and for the fix. The example is probably a bit verbose for this doc, but not terrible.

PR #187, opened 2022-03-06
Patch spec

warpfork:
Introducing a "patch" system, and defining it by fixtures!

This is roughly following the [JSON Patch (RFC 6902)](https://datatracker.ietf.org/doc/html/rfc6902/) specification -- It's a very decent specification.  (Most of the fixtures here also come from its examples appendix, sometimes with minor modifications.)

In brief highlights:
- There are six operations -- add, replace, remove, copy, move, test.
  - (This is the same as RFC 6902.)
- Operations are supplied in a list, and applied linearly.  Any operation erroring results in the entire patch not being applied.
  - (This is the same as RFC 6902.)
- Operations do not create parent paths automatically.
  - (This is the same as RFC 6902.)
  - (I think we may want to add ways to support this succinctly in the future!)
- Note there is no "upsert" operation.
  - (This is the same as RFC 6902.)
  - (I think we may want to add ways to support this succinctly in the future!)
- All of these fixtures are JSON -- but of course, because this is IPLD, both the subject data, as well as the patch instructions themselves, can be communicated in any codec (e.g. CBOR, or dag-pb (... for the subject data anyway), or etc).
- Note that we do *not* follow RFC 6902's [escaping rules](https://datatracker.ietf.org/doc/html/rfc6902/#appendix-A.14) (which appear to be from [RFC 6901](https://datatracker.ietf.org/doc/html/rfc6901#section-3), although this is not explicitly stated).  The opinion among us who did an early design review for this system was that the escaping mechanism shown there is very unusual, and would surprise most users, and thus we will not replicate it.
- Note that a Patch system in IPLD is expected to maintain order whenever possible.  Wherever this is ambiguous (e.g. operations adding data to a map under a key that didn't previously exist), operations are defined as appending to the end of the relevant node.  This results in some of our fixtures being mildly divergent from RFC 6902 (which appears to treat map order as inconsequential).  We consider this important to specify for determinism reasons.

There is a matching implementation, which exercises these fixtures, found in https://github.com/ipld/go-ipld-prime/pull/350 .

### DIFF
--- a/specs/patch/fixtures/fixtures-1.md
+++ b/specs/patch/fixtures/fixtures-1.md
@@ -1,0 +1,89 @@
+Patch Fixtures 1
+================
+
+This file contains a bunch of fixtures for the Patch system.
+These are suitable both as mechanical test fixtures, and as examples for humans.
+
+### What's covered?
+
+All the fixtures in this file are operating on single blocks --
+the data can still be a tree (maps, lists, etc), but contains no links.
+(Other fixtures cover data that spans multiple blocks by use of links.)
+
+### What's the fixture format?
+
+This is the [testmark format](https://github.com/warpfork/go-testmark);
+it can be consumed programmatically.
+(Note that you probably _can't_ see these labels as this page is rendered on ts.
+
+### This doesn't have to be JSON!
+
+We've used JSON here for the initial data, as well as the patch instructions and the results.
+However, this is a totally stylistic choice.  We're just using JSON here because it's convenient and it's human-readable.
+You can equally well express all of these objects in CBOR, or other [codecs](/docs/codecs/); it makes no difference to IPLD.
+The patch instruction and the subject data don't even have to be in the same codec, either.
+
+
+Fixtures
+--------
+
+### Adding an entry to a map
+
+Given an initial document:
+
+[testmark]:# (adding-a-map-entry/initial)
+```json
+{"foo": "bar"}
+```
+   
+and a Patch OperationSequence:
+
+[testmark]:# (adding-a-map-entry/patch)
+```json
+[
+	{ "op": "add", "path": "/baz", "value": "qux" }
+]
+```
+
+The following document should result:
+
+[testmark]:# (adding-a-map-entry/result)
+```json
+{
+	"foo": "bar",
+	"baz": "qux"
+}
+```
+
+Note that if order is sensitive, the additions are considered to go at the end of map.
+
+
+### Inserting into a list
+
+Given an initial document:
+
+[testmark]:# (inserting-into-a-list/initial)
+```json
+["bar", "baz"]
+```
+   
+and a Patch OperationSequence:
+
+[testmark]:# (inserting-into-a-list/patch)
+```json
+[
+	{ "op": "add", "path": "/1", "value": "qux" }
+]
+```
+
+The following document should result:
+
+[testmark]:# (inserting-into-a-list/result)
+```json[
+[
+	"bar",
+	"qux",
+	"baz"
+]
+```
+
--- a/specs/patch/fixtures/fixtures-1.md
+++ b/specs/patch/fixtures/fixtures-1.md
@@ -281,3 +281,45 @@ The following document should result:
 	"bar": "zar"
 }
 ```
+
+
+### Testing and conditional modification - Failing case
+
+Given an initial document:
+
+[testmark]:# (test-and-conditional-fail/initial)
+```json
+{
+	"baz": "qux",
+	"foo": [
+		"a",
+		2,
+		"c"
+	]
+}
+```
+
+and a Patch OperationSequence:
+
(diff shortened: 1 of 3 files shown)


warpfork:
I think this is mergable.  I'd like to add even more fixtures in the future (especially, some that examine how scenarios unfold when a patch fails to apply), and expand the system to include whole instruction message formats for applying patches to documents starting at a CID, and so forth -- but these fixtures here so far should all remain pretty incontrovertable even as we continue expanding the system, I think.

RangerMauve (commented):
This looks really handy!

I think it'd be nice if the examples contained one or two more complex structures like adding a  `Link` to a UnixFS directory.

Also, do the operations work with other Kinds like `Bytes`? Can I `delete` a single byte from the middle?

specs/patch/fixtures/fixtures-1.md
```diff
@@ -0,0 +1,283 @@
...
+This is the [testmark format](https://github.com/warpfork/go-testmark);
+it can be consumed programmatically.
+(Note that you probably _can't_ see these labels as this page is rendered on the website;
+you'll have to find the markdown [source](/specs/about/#source).)
+
+### This doesn't have to be JSON!
+
+We've used JSON here for the initial data, as well as the patch instructions and the results.
+However, this is a totally stylistic choice.  We're just using JSON here because it's convenient and it's human-readable.
+You can equally well express all of these objects in CBOR, or other [codecs](/docs/codecs/); it makes no difference to IPLD.
```
mvdan, line 24: nit: shouldn't you call these DAG-JSON and DAG-CBOR?

mvdan (approved):

specs/patch/fixtures/fixtures-1.md
warpfork: I'm honestly not clear on if that's more helpful or less.

I do actually parse these with the json codec and not the dag-json codec in the first code I wrote using them, iirc.  But I suppose that's going to change the instant I add a fixture that crosses blocks!

mvdan: Well, to the rest of the world, JSON and CBOR are just the regular standards, without the detersministic encoding quailties that the DAG forms bring with them. Even if links are not involved, you probably do not mean just plain old JSON/CBOR.

warpfork:
> I think it'd be nice if the examples contained one or two more complex structures like adding a Link to a UnixFS directory.

Yeah, I definitely want multi-block fixtures.

I dunno about UnixFS; it would be nice, but that hits a deeper dependency tree.  (Some other PRs have done this recently, but it prompted a discussion already about practical maintainability, and I dunno if we have ideal answers yet.)  We also don't necessarily want to suggest that that's a trivial thing that people should be doing bare; ideally that goes through an ADL code path so that sharding works transparently.  (We've had bad experiences in the wild in the past with UnixFS directories being patched until they become too large to fit in the transfer size limit for a block; whoops.)

> Also, do the operations work with other Kinds like Bytes? Can I delete a single byte from the middle?

Not yet.  I think we'd have to introduce new operations for that.

The JSON Patch similarly doesn't specify such a feature for operating on strings.

We certainly could add another op code for that.  But I probably won't try to heap that into this first pass.

BigLep:
This was fun to read.  Thanks @warpfork .  It looks like you got approval.  Are we good to merge?

Also, for someone validating their implementation, do we need to provide more intricate structures (more nestings, maps inside lists, etc.)?

specs/patch/fixtures/fixtures-1.md
```diff
@@ -0,0 +1,283 @@
...
+```json
+[
+	{ "op": "test", "path": "/baz", "value": "qux" },
+	{ "op": "test", "path": "/foo/1", "value": 2 },
+	{ "op": "add",  "path": "/bar", "value": "zar" }
+]
+```
+
+(Note that this contains both test operations as well as a subsequent add operation!
+The add operation should only apply if the test operations yield true.)
```
rvagg, line 268: this note suggests the need for a follow-up test case where it doesn't yield true, to make a test runner actually test `test` operations and not just the `add`

I reckon just copy and paste these 3 but change the second `test` value to `3` and then the last fixture shouldn't have `bar`

specs/patch/fixtures/fixtures-1.md
RangerMauve: Added a mention that it's DAG-JSON and DAG-CBOR

specs/patch/fixtures/fixtures-1.md
RangerMauve: I tried adding this and the result is a failing test case with `patch_test.go:63: patch did not apply: test failed`. Does testmark have a "should fail" flag of some sort?

RangerMauve: Gonna add a thing where if the test name ends in `-fail`, it'll be expected to fail.

PR #188, opened 2022-03-16
Add reify to glossary

RangerMauve:
This is based on the docs for the Reify function in the Golang ADL spec and from talks with @warpfork


rvagg (approved):
sgtm although this arises from @warpfork's particular take on the word so it might be best for him to give a :thumbsup:

warpfork (approved):
LGTM too!  Let's ship it

PR #190, opened 2022-03-21
exploration report: lens combinators

warpfork:
This is a brief exploration of some ideas about how we could declaratively specify the use of Schemas, ADLs, and even have an extension path to other as-yet-uninvented data transformation mechanisms.

In it, Schemas and ADLs are both reframed as "lenses", and another new "lens" is also proposed: a simple "lens combinator", which allows invoking other lenses at various paths within some data.

Examples of using it as part of a Patch API are included, but the concept is meant to be reusable and drop-in in other scenarios as well.

### DIFF
--- a/notebook/exploration-reports/2022.03-lens-combinators.md
+++ b/notebook/exploration-reports/2022.03-lens-combinators.md
@@ -1,0 +1,164 @@
+Lens Combinators: a proposal for extensible declarative signalling of advanced IPLD features
+============================================================================================
+
+**Author**: Eric Myhre ([@warpfork])
+
+---
+
+A brief example of why there's something to do here:
+
+Here's an instruction I wish we could give to a patch API:
+
+First, using just one ADL:
+
+```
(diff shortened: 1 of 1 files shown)


RangerMauve:
I like that the lens concept makes it explicit that there's different ways to interpret data and gives clear signals on what those ways are. Being able to apply lenses to different paths is really powerful.

Funky ideas for URLs:

```
# ADL at the top level
?adl=HAMTv1
# ADL at specific path
?adl-{urlEncode(path)}=HAMTv1
# schema for the whole tree
?schema={urlEncode(ipns://bafywhatever#startypeNameHere)}
# Schema for a specific path
?schema-{urlEncode(path)}={urlEncode(ipns://bafywhatever#startTypeNameHere)}
# A bunch in a queryselector
?adl=HAMTv1&adl=SomethingElse&schema-urlEncode(/foo/bar)=urlEncode(ipns://bafywhatever
```

PR #191, opened 2022-03-21
Update what-is-ipld image

RangerMauve:
Based on @warpfork 's latest graphic


RangerMauve:
For some reason the `.jpg` file isn't being found. 🙃

RangerMauve:
Should be good to go. 👍

BigLep:
Is there discussion somewhere that discusses what was wrong with the old graphic?

RangerMauve:
Not sure if there's specific places that discuss this, but @warpfork was thinking of remaking it with some new changes.

The main difference is that the concept of "Universal Interface" is centered and the Data Model is a subset of that rather than the old image combining the concept of data model with Universal Interface.

warpfork:
Previous generation of discussion that produced the standing what-is-ipld.png image was here: https://github.com/ipld/specs/issues/244

I think even at the time, we were pretty aware that we were having a tough time figuring out how to draw the raw (or "bare", I guess some posts back then called it?  same thing) Data Model in order to make some of these relationships clear.

At the same time, the existence of a universal interface that (esp.) pathing and other features like Selectors can be built over -- and that that interface abstracts over not just codecs, but also over things like ADLs -- is a really huge important thing to pitch.  Transparent pathing over sharding is just a killer feature, and it sucks if people don't realize IPLD has that capability!

Since shipping the last generation of diagram, anecdotally, I've found we do indeed still struggle to get that communicated to people through the docs website and the existing diagram alone.  Usually it gets across in chat messages, if we kept someone's attention long enough for them to come ask about it.  Great when a chat like that happens, but it would be a lot more ideal if that piece of info propagated itself better without that much energy required.

The hope is that this diagram might nail it better.

warpfork:
(We may want to open a design shop request to see if we can get a prettier version of it, before shipping it to the website, though.)

warpfork:
@rvagg you'd also be a very wonderful set of eyes to have visit this, if you have the time.

rvagg:
Oh, so, this new graphic is OK in terms of what it conveys, I like the arrows and flow, but we can't replace a profesh graphic with a whiteboard photo. We're also removing content as the old one had explanatory text which was helpful, this one has less text, which might be OK as long as we're adding explanatory notes in the textual context on the page. Basically I'm unsure if this is additive or subtractive! It seems to have ticks in both columns.

BigLep:
> (We may want to open a design shop request to see if we can get a prettier version of it, before shipping it to the website, though.)

I think that's a good idea.

warpfork:
Okay.  Agree on all counts.  The good news is: the issues that are currently subtractive sound fixable.  (Agree that we'll need more text on the side.  (Although fwiw, I think having it out of the image will actually make it a lot easier to lay out the text!  The current image is pretty mega.))

Opened a request in the design shop to professionalize the graphic: https://github.com/protocol/design-shop/issues/1388

(For copious crosslinking: https://github.com/protocol/design-shop/issues/306 was the design-shop issue for the last generation of this, about two years ago.)

RangerMauve:
Should I add the text from the previous image to the body of the page as part of this PR?

BigLep:
@RangerMauve : it looks like we're unblocked in https://github.com/protocol/design-shop/issues/1388 .  Feel free to move out of draft when this is ready for @rvagg to review.

RangerMauve:
Added the image from the design shop instead of the whiteboard sketch.

Also added descriptors for the various boxes in the image based on what was within the image before and some stuff I pulled out of thin air (other docs).

Wording and spelling corrections would be very much appreciated.

RangerMauve:
Should I make the descriptors for the boxes into links that users could follow for more info?

rvagg:
As per the comment I left in the design-shop thread, I'm not super keen on including **"Lens combinators"** because it's theoretical, there's nothing of substance behind this that we can point to when people come along asking for it aside from an exploration report. We've had to shoulder a lot of criticism over the last few years about the theoretical nature of so much of the IPLD stack; I was even hearing some fresh criticism of IPLD today (!) that was along these lines so I'd really really like to focus on the practical and the usable with the messaging that we frontload, lest we continue to have IPLD written off as a research project that can be side-stepped for practical work.

When I show this to Launchpad participants, like I did a couple of weeks ago (the whiteboard image), I have to include comment that this thing isn't something you can use yet, but it's an _idea_ that _may_ come to fruition when someone invests the time in it.

So my preference is for the image without "Lens combinators", but with the option to add it back in if usable code appears. But, if you feel strongly that it should be included, then I'd really like a caveat of some kind in the description next to it saying that it's "theoretical" or just a "planned feature".

The other two problematic ones that are slightly similar are ADLs and Patch. You can use ADLs today, but the write interface is very experimental and there's much more to do before they have the kind of usability that something like Selectors do. Even less mature is Patch, which is still just a [pull request](https://github.com/ipld/go-ipld-prime/pull/350), which is at least more than "Lens combinators" which is just an exploration report. Perhaps a similar disclaimer re maturity would be good along with these too. "Under development" or something similar?

RangerMauve:
> As per the comment I left in the design-shop thread, I'm not super keen on including **"Lens combinators"** because it's theoretical, there's nothing of substance behind this that we can point to when people come along asking for it aside from an exploration report.

That's fair. I think it'd be safe to put the Lens Combinator portion of the image onto the backlog until there's some APIs to support it. We have the image now so it wouldn't be too hard to add once it's ready. 👍

> The other two problematic ones that are slightly similar are ADLs and Patch... Perhaps a similar disclaimer re maturity would be good along with these too. "Under development" or something similar?

A disclaimer sounds fair. Should I link to the Patch PR?

RangerMauve:
Updated per @rvagg 's comments, changed the image to an SVG.

Should be good to go? (warning: I probably have a spelling error in there somewhere. 🙃)

RangerMauve:
I used mdspell to find some spelling errors and fixed them. 😁

docs/intro/hello-world.md
```diff
@@ -60,11 +60,23 @@ and [Advanced Data Layouts](/docs/advanced-data-layouts/), which let us do thing
...
+- `Advanced Data Layouts` couple programmatic functionality with data structures, allowing for complex data structuring while presenting as the Data Model (e.g. encryption, large multi-block spanning data structures). Note that ADL APIs are more stable for reading than writing as write APIs are under development.
+- `Codecs` are bound by the data model to include a set of data types, including links (currently represented as CIDs). Codecs supporting a subset of the Data Model (DAG-PB, plain CBOR, plain JSON, etc.) likewise have limited support from functionality from Data Model components.
+- `Blocks` are arbitrary arrays of bytes identified by a CID (content identifier, including hash and codec details). IPLD doesn't concern itself with the source or nature of these bytes as long as its Codecs can read and/or write them. Limitations (size, location, availability, etc.) are concerns of the data source.
+- The `Universal Interface` the interface that applications use which abstracts over top of the raw data model and lenses like Schemas and ADLs. It can then be used by higher level APIs to load and manipulate data.
```
rvagg, line 73:
```suggestion
- The `Universal Interface` is the interface that applications use which abstracts over top of the raw data model and lenses like Schemas and ADLs. It can then be used by higher level APIs to load and manipulate data.
```

rvagg (approved):
lgtm aside from the one small omission

PR #195, opened 2022-03-29
Intiial exploration report for IPLD URL Scheme

RangerMauve:
Starting a PR so that folks can start bikeshedding early on.

Not ready for review until I fill out a bunch of TODOs.

### DIFF
--- a/notebook/exploration-reports/2022.03-ipld-url-scheme.md
+++ b/notebook/exploration-reports/2022.03-ipld-url-scheme.md
@@ -1,0 +1,106 @@
+IPLD URL Scheme v0
+==================
+
+This is a proposed IPLD URL scheme spec, focusing on a minimap viable product which doesn't address advanced use cases such as lenses or selectors.
+
+**Author**: Mauve Signeaver ([@RangerMauve])
+
+---
+
+tl;dr
+
+```
+GET ipld://bafy....etc/some/path => {some: data}
+PUT ipld://bafy....etc/some/path {some: data} => ipld://bafy...etc/
+```
+
+## Objective
+
+`ipfs://` exists out in the wild to provide ways to link to files, and some web browsers integrate IPFS URLs to load and author content.
+This spec shows how a similar URL scheme can exist for IPLD data.
+Part of the goal of the spec is to make it easy to reuse more or less the same APIs between software that handles `ipld://` URLs natively, HTTP gateways like the [IPFS gateway](https://docs.ipfs.io/concepts/ipfs-gateway/) and CLIs.
+With that in mind, there should be a clear mapping between HTTP-like APIs via gateways and URL handlers, and CLI arguments.
+In order to make it easier to come to an initial spec, we will focus on a minimal set of features and make subsequent specs with support for features like ADLs, Schemas, and advanced selectors.
+
+## Initial Functionality
+
+The initial functionality focuses on being able to GET and PUT IPLD data to and from a URL.
+For now, this will be done by encoding individual blocks of IPLD data using codecs with a default to the [DAG-JSON](https://ipld.io/specs/codecs/dag-json/spec/) codec.
+Some initial support for pathing is also included for convenience.
+
+### GET `ipld://{CID}/{Path} Headers {Accept="application/vnd.ipld.dag-json"} => Data`
+
+The most basic mapping is to get some data from a CID.
+
+The `CID` parameter in the URL MUST be in the CIDv1 format or any other format that is not case sensitive.
+This is necessary since the URL scheme considers the `hostname` portion to be case-insensitive and many URL parsers (such as the one in Chrome) will automatically convert it to lowercase. 
+
+`Data` is the response body which contaisn the encoded IPLD data for the given CID+Path.
+Note that any `Link`s within the IPLD data will not be resolved, and you'll need to conver them to URLs and fetch additional data yourself, or add path segments to the URL to traverse into them.
+
+The `Accept` header may be optionally provided to specify the codec that should be used in the response.
+This follows the same conventions as in the [IPFS gateway](https://github.com/ipfs/go-ipfs/pull/8758).
+Other codecs that may be targetted are DAG-CBOR.
+TODO: `application/vnd.ipld.dag-json` or just `application/json`?
+
+The `Path` can be used to specify segments separated by a `/` to use with basic traversal,
+if it is supplied the IPLD Node at that location relative to the CID will be returned.
+If there path does not lead to a valid IPLD node, an error will be returned instead, e.g. `HTTP 404`.
+
+### PUT `ipld://{CID}/{path} Headers {Content-Type="application/vnd.ipld.dag-json"} Body={encoded IPLD data} => URL`
+
+The `{CID}` should be the "base" CID to extend data from.
+Sadly if we are to use URLs (TODO: why should we use URLs?), we must have some sort of CID in the hostname.
+TODO: Document some inline CIDs to use. IPFS has this empty directory CID `ipfs://bafyaabakaieac` which would be cool to have analogs for for empty `{}` object and `[]` array nodes.
+
+The `Body` of the request should contain the encoded IPLD data you wish to upload.
+
+The `Content-Type` may be optionally provided to specify the encoding of the `Body`.
+By default it will be assumed to be `DAG-JSON`.
+
+If a `Path` is specified, intermediate Nodes will be created and added to the Node specified by the CID.
+The `Body` will be added as the last node in the chain. TODO: Wording?
+If data already exists at intermediate nodes, it won't be touched other than the final part of the segment.
+
+The response will contain an `ipld://` URL which can then be used with `GET` or subsequent `PUT`s.
+
+### URL to Gateway conversion:
+
+`ipld://` URLs can be converted to standard Gateway URLs by replacing the `ipld://` portion at the beginning with `http://localhost:8080/ipld/` and passing the rest of the request headers and body as is.
+Note that gateways may have different root paths than `http://localhost:8080/ipld/`, e.g. `https://ipld.io/ipld/`.
+
+E.g.: `ipld://{CID}/{Path}` => `http://localhost:8080/ipld/{CID}/{Path}`
+
+### URL to CLI
+
+There's no planned CLI for this functionality yet, but a request to a URL might translate to one using something like this:
+
+```
+PUT ipld://{CID}/ Headers {Content-Type="application/vnd.ipld.dag-json"} Body="{}"
+
+echo "{}" | ipld-cli put ipld://{CID}/ --content-type="application/vnd.ipld.dag-json"
+
+Or
+
+GET ipld://{CID}/{Path} {Accept="application/vnd.ipld.dag-json} => Data
+
+ipld-cli get ipld://{CID}/{Path} --accept="application/vnd.ipld.dag-json" > myfile.json
+```
+
+## Schema
+
+The operations are documented via an IPLD Schema.
+
+```ipld=
+// TODO: Is this necessary for v0?
+```
+
+## Gotchas
+
+- In order to support `/` inside path segments, it should be escaped using [URI component escaping](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent). This could potentially be used for encoding binary data inside paths, too.
+
+## Future work:
+
+- Lenses: ADLs/Schemas
+- `PATCH` support
+- Selectors


notebook/exploration-reports/2022.03-ipld-url-scheme.md
```diff
@@ -0,0 +1,106 @@
...
+The `CID` parameter in the URL MUST be in the CIDv1 format or any other format that is not case sensitive.
+This is necessary since the URL scheme considers the `hostname` portion to be case-insensitive and many URL parsers (such as the one in Chrome) will automatically convert it to lowercase. 
+
+`Data` is the response body which contaisn the encoded IPLD data for the given CID+Path.
+Note that any `Link`s within the IPLD data will not be resolved, and you'll need to conver them to URLs and fetch additional data yourself, or add path segments to the URL to traverse into them.
+
+The `Accept` header may be optionally provided to specify the codec that should be used in the response.
+This follows the same conventions as in the [IPFS gateway](https://github.com/ipfs/go-ipfs/pull/8758).
+Other codecs that may be targetted are DAG-CBOR.
+TODO: `application/vnd.ipld.dag-json` or just `application/json`?
```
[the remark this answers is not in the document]
RangerMauve: My only worry with regards to using `application/vnd.ipld.dag-json` is that it wouldn't get recognized by common browser tools like syntax highlighting extensions and the such. I think it could still work with `await (await fetch(url)).json()` in browsers, though.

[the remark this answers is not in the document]
RangerMauve: My only worry with regards to using `application/vnd.ipld.dag-json` is that it wouldn't get recognized by common browser tools like syntax highlighting extensions and the such. I think it could still work with `await (await fetch(url)).json()` in browsers, though.

notebook/exploration-reports/2022.03-ipld-url-scheme.md
```diff
@@ -0,0 +1,106 @@
...
+
+The `Path` can be used to specify segments separated by a `/` to use with basic traversal,
+if it is supplied the IPLD Node at that location relative to the CID will be returned.
+If there path does not lead to a valid IPLD node, an error will be returned instead, e.g. `HTTP 404`.
+
+### PUT `ipld://{CID}/{path} Headers {Content-Type="application/vnd.ipld.dag-json"} Body={encoded IPLD data} => URL`
+
+The `{CID}` should be the "base" CID to extend data from.
+Sadly if we are to use URLs (TODO: why should we use URLs?), we must have some sort of CID in the hostname.
+TODO: Document some inline CIDs to use. IPFS has this empty directory CID `ipfs://bafyaabakaieac` which would be cool to have analogs for for empty `{}` object and `[]` array nodes.
```
rvagg, line 54: Well, we now have the concept of "unit" type in schema-land: https://github.com/ipld/ipld/blob/cfe4b4914b6c5a6e37c5cc9741528e0022337f98/specs/schemas/schema-schema.ipldsch#L761-L789

And we have identity CIDs, we could just pick one to standardise on as a base.

https://cid.ipfs.io/#baeaaaapw might be a good choice, `null` is a commonly understood unit value and might be a good "there's nothing here to start from".

Or are you wanting to convey codec information with the "base"? So, Content-Type tells the receiver what format the data is being provided in, but the base CID says what codec should be used for the encoded block? Do we want that kind of flexibility? And if so, what about base CIDs that have some data because they're CIDs that reference some existing block or have some data in them as identity CIDs?

Yet another possibility here is to just use identity CIDs to send the data and `Content-Type` to say what codec to encode into—what limitations do we run into with hostnames here?

notebook/exploration-reports/2022.03-ipld-url-scheme.md
```diff
@@ -0,0 +1,106 @@
...
+
+The `Body` of the request should contain the encoded IPLD data you wish to upload.
+
+The `Content-Type` may be optionally provided to specify the encoding of the `Body`.
+By default it will be assumed to be `DAG-JSON`.
+
+If a `Path` is specified, intermediate Nodes will be created and added to the Node specified by the CID.
```
rvagg, line 61: Does this strictly stick to the integer components of a path, where if it encounters a `0` it'll create an array for you? And what if the path has a `1` (or later) in it? Do we make an array that has that many elements that are ... empty? Because we'd have to know what "empty" is (back to the _unit_ discussion).

notebook/exploration-reports/2022.03-ipld-url-scheme.md
rvagg: Also, maybe the semantics of a `PUT` could all fall under IPLD Patch since it already has all needed modification primitives (and more)? Otherwise this would be introducing a mutation operation that is limited to only adding or changing data from existing blocks, whereas Patch gives us full modification potential.

notebook/exploration-reports/2022.03-ipld-url-scheme.md
```diff
@@ -0,0 +1,106 @@
...
+The `Body` will be added as the last node in the chain. TODO: Wording?
```
rvagg, line 62: yeah, avoid chain, "tree" or "graph" or even just "block" might be best here.

rvagg:
This steps on the `/ipld/` convention that was somewhat formalised in the go-ipfs v0.10 release in the DAG API, where it's not an explicit fallback to IPLD pathing, avoiding the UnixFS magic that has got in the way previously. `/ipld/` was an existing prefix but now it's a specific tool to opt for pure IPLD pathing. I think/thought there was moves to do this on the gateway too, which is the same as the `/ipld/` CID prefix suggested here.

That may not be a conflict, however, `ipfs dag get /ipld/bafy.../foo/bar/baz` should be the same as here, `ipfs dag put ...` doesn't have the concept of a "base" but could probably be made to do so, thereby enabling (roughly?) the same PUT operation being proposed here.

Either way, it might be worth looking at that and making sure we're not doubling up on something that's not _quite_ the same and may end up being confusing to users. If there's a path to harmonisation then that'd be awesome.

notebook/exploration-reports/2022.03-ipld-url-scheme.md
RangerMauve: > Or are you wanting to convey codec information with the "base"?

I think keeping the codec information in the `Content-Type` header would be best to keep things URL-y. `ipld://baeaaaapw/` seems like a great place to start.

> what about base CIDs that have some data because they're CIDs that reference some existing block or have some data in them as identity CIDs?

I think this would be useful. e.g. It'd be cool to have some base CIDs with examples like an empty array or an empty map. Is there an existing list I could reference somewhere? Should I just generate these myself with the ipld CLI?
'

> maybe the semantics of a PUT could all fall under IPLD Patch since it already has all needed modification primitives

I was actually thinking that the `PATCH` HTTP method could be used there instead. I think it'd be cool to add support for it as a first-class feature, but since Patch is still a bit of a WIP I didn't want to block this to wait on it.

One thing to consider here is that we just need to have _something_ in the hostname portion of the URL since in my experience most parsers (like browsers) will freak out if it's only _sometimes_ there. 😅

Alternately, in IPNS land we were looking at using `localhost` as a "no-op" name for when we want to do something with data that isn't already tied to a particular CID. @lidel might have more to say on that front.

notebook/exploration-reports/2022.03-ipld-url-scheme.md
RangerMauve: I think to be safe it would be good to generate intermediate Map Nodes if a node doesn't already exist at that path otherwise it might be too magical. 😅 E.g. what if somebody is iterating through a list of names and one of them just happens to be an integer.

Also, should we error out if somebody tries to add a named key to a node which is an array?

RangerMauve: I think this could also be augmented down the line with Schemas so you can give better type hints for individual segments of the path.

RangerMauve: Yeah, I think erroring out when there isn't something obvious and having fancier behavior in the future would be best for this initial version. I'll add some notes about how lists should be handled in the meantime though.

RangerMauve:
@lidel Would you be able to comment on [@rvag's comment](https://github.com/ipld/ipld/pull/195#issuecomment-1103417741) regarding conflicts with what is already in the IPFS gateway?

My gut feeling is that the `ipfs dag` API wouldn't be affected by these changes since we would likely be exposing it on the gateway API instead.

notebook/exploration-reports/2022.03-ipld-url-scheme.md
```diff
@@ -0,0 +1,111 @@
...
+PUT ipld://bafy....etc/some/path {some: data} => ipld://bafy...etc/
+```
+
+## Objective
+
+`ipfs://` exists out in the wild to provide ways to link to files, and some web browsers integrate IPFS URLs to load and author content.
+This spec shows how a similar URL scheme can exist for IPLD data.
+Part of the goal of the spec is to make it easy to reuse more or less the same APIs between software that handles `ipld://` URLs natively, HTTP gateways like the [IPFS gateway](https://docs.ipfs.io/concepts/ipfs-gateway/) and CLIs.
+With that in mind, there should be a clear mapping between HTTP-like APIs via gateways and URL handlers, and CLI arguments.
+In order to make it easier to come to an initial spec, we will focus on a minimal set of features and make subsequent specs with support for features like ADLs, Schemas, and advanced selectors.
```
aschmahmann, line 23: While I get the temptation to just do this a little bit at a time, I think we should do it with our eyes towards the future.

In particular, my concern is that `ipld://` as described is almost exactly the same as `ipfs://` and has limited utility in that there doesn't seem to be a way to include ADLs or selectors in a way that wouldn't also work for `ipfs://`. This means that what we get out of `ipld://` in comparison to `ipfs://` is a fairly limited domain (and is what @rvagg alluded to with how `/ipld` currently works in some go-ipfs commands) that AFAICT mostly boils down to debugging UnixFS (at least on the GET side). So why bother asking implementers (e.g. browsers) to support both `ipfs://` and `ipld://`?

In a recent experiment with using selectors with `ipfs://` I went with `?selector=` which also allowed me to utilize ADLs (code starts here https://github.com/aschmahmann/go-ipfs/blob/b0859a906110bdac348d2888dfd6ddd960df738e/core/corehttp/gateway_handler.go#L388). `ipfs://<cid>?selector=<selector>?format=<some format>` seems like it would cover almost all of the GET use case described here for `ipld://`.

I wonder if we could help ourselves out here with some type of escaping in `ipld://` that will allow us to do more interesting things that we can't in `ipfs://` for legacy reasons (i.e. no reserved characters). For example: `ipld://bafyfoobar/[ADL=MyHAMTv1]/field1/[ADL=YourHAMTv2]/field2/[ADL=FBL]/[Selector=Slice[100,400]]` might be more readable than some base-encoded inline selector or CID of a selector for the cost of disallowing `[...]` as path segments (or require escaping in that case)

RangerMauve: That's a really good point regarding the existing IPFS gateway APIs already being able to support most of the `GET` use cases. I was imagining that that would change more as time went one. I think the authoring of IPLD data with `PUT` was more interesting to me as a differentiator since the writable gateway APIs are currently very file-focused and it'd feel super hacky to add a new mode to them for IPLD with `?format`.

Authoring IPLD data as well as consuming it with IPLD-specific semantics is the main motivator for having this be a separate URL scheme / gateway path for me.

I really like the idea of having extra parameters within individual segments of the path!
I think this could also help with the stuff @rvagg mentioned about type hints for individual nodes that get auto-created when you're using `PUT` to add a node to a given path.

From your example URL, the format seems to be `ipld://CID/SEGMENT_OR_SELECTORS`, but what if it we added the selectors onto the segment itself?

e.g. `PUT ipld://bafyduck/field1[Kind=Map]/field2`?

rvagg: _If_ you go with special characters, keep in mind possible escaping options--we face critique in a few of our specs for the lack of escaping to allow the full domain of fields/names/etc. Pathing is particularly problematic in this respect since it is already restricted in the domain of field names. Taking `[` & `]` off the board as well would be unfortunate (although not the worst thing, but we do tend to see a bit of `[`, `]` use for field names in data mainly thanks to web tooling conventions). I don't really want to see this descend into a rabbit hole of optionality but a bit of forward thinking might be helpful.

BigLep:
@RangerMauve : do you want to incorporate any more feedback here, or are you good if we merge now?

RangerMauve:
@BigLep Gonna go over this weekend to see if there are more notes and TODOs to add and then I think I'll be ready to merge next week. 👍

RangerMauve:
Based on the discussion we had at the end of the last IPLD weekly meeting, I've added a section on escaping special characters and reserved some future syntax for Lenses based on the suggestions from @aschmahmann 

One thing I'd like resolved before we merge is to get some more example unit CIDs to base off of.

Ideally it'd be nice to have the following:
- `null`
- `{}` empty Map
- `[]` empty List

Would somebody mind linking me to info about how I may generate these or find an existing set? The encoding doesn't matter too much, but DAG-JSON might be easiest to mess with manually.

rvagg:
Here's how I make these things, I have the codecs and js-multiformats repos locally and built because I need to `require()` from the dist/ directories in the repl, but you can do the same from npm installed versions since those are packaged dist/ directories.

```
> var { CID } = require('./js-multiformats/dist/')
> var dagJSON = require('./js-dag-json/dist')
> var dagCBOR = require('./js-dag-cbor/dist')
> var { identity } = require('./js-multiformats/dist/hashes/identity')
> var raw = require('./js-multiformats/dist/codecs/raw')
> var digest = require('./js-multiformats/dist/hashes/digest')
> CID.create(1, raw.code, digest.create(identity.code, new Uint8Array([0])))
bafkqaaia
> CID.create(1, raw.code, digest.create(identity.code, new Uint8Array([])))
bafkqaaa
> CID.create(1, dagJSON.code, digest.create(identity.code, dagJSON.encode(true)))
baguqeaaeorzhkzi
> CID.create(1, dagJSON.code, digest.create(identity.code, dagJSON.encode([])))
baguqeaaclnoq
> CID.create(1, dagJSON.code, digest.create(identity.code, dagJSON.encode({})))
CID(baguqeaacpn6q)
> CID.create(1, dagJSON.code, digest.create(identity.code, dagJSON.encode('')))
CID(baguqeaaceira)
> CID.create(1, dagCBOR.code, digest.create(identity.code, dagCBOR.encode(true)))
bafyqaapv
> CID.create(1, dagCBOR.code, digest.create(identity.code, dagCBOR.encode([])))
bafyqaama
> CID.create(1, dagCBOR.code, digest.create(identity.code, dagCBOR.encode({})))
bafyqaana
> CID.create(1, dagCBOR.code, digest.create(identity.code, dagCBOR.encode('')))
CID(bafyqaala)
> CID.create(1, dagJSON.code, digest.create(identity.code, dagJSON.encode('banananana, do do de do do')))
CID(baguqeaa4ejrgc3tbnzqw4ylomewcazdpebsg6idemuqgi3zamrxse)
```

notebook/exploration-reports/2022.03-ipld-url-scheme.md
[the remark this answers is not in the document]
lidel: @RangerMauve fair point! so.. `application/json` everywhere + also support `application/vnd.ipld.dag-json` in requests? that should maximize reuse of things that have nice UX/support for JSON.

notebook/exploration-reports/2022.03-ipld-url-scheme.md
lidel: - I think my idea was that if we want to invent some name to indicate "empty" state, the `localhost`  is not a bad choice because it is something devs understand, resolves safely to loopback device if DNS is ever involved (either on purpose, or as a bug),  and also hints that data is added to local IPLD node, and not some magical cloud:
  - `PUT ipld://localhost`  → get CID

- In case a drive-by reader is confused by "JSON Patch", it is different  thing from "HTTP PATCH" method:
   - The "JSON Patch"   @rvagg mentioned is a convention ([RFC 6902](http://tools.ietf.org/html/rfc6902) from the IETF) for representing JSON diffs (http://jsonpatch.com) – see relevant example in https://github.com/ipfs/go-ipfs/issues/4801#issuecomment-1024303462  (and the discussion that follows).

lidel:
@RangerMauve  would be useful to add any missing edge cases to https://github.com/ipld/codec-fixtures  (if not present already) – we point more and more people at fixtures like these, so it is good return of time invested.

notebook/exploration-reports/2022.03-ipld-url-scheme.md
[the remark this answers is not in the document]
RangerMauve: Yeah, I'm into that. I'll add it to the doc. 😁

notebook/exploration-reports/2022.03-ipld-url-scheme.md
RangerMauve: Thinking more on this, maybe we should error out when there's no intermediate node? E.g., users would need to manually add intermediate nodes with separate requests as they're building up the DAG.

notebook/exploration-reports/2022.03-ipld-url-scheme.md
RangerMauve: I'm into having `ipld://localhost/` be a special thing for when there's no base CID to build off of. @rvagg @warpfork Any strong feelings on that matter?

Regarding PATCH, I've added a link to the GH issue where the implementation is being tracked: https://github.com/ipld/go-ipld-prime/pull/350

rvagg: seems reasonable to me; it is slightly arbitrary (e.g. why not `::` to be uber-modern?) but I guess anything we choose is!

notebook/exploration-reports/2022.03-ipld-url-scheme.md
RangerMauve: @rvagg Mind weighing in on whether we should error out when there's no intermediate node?

rvagg: Hmmmm, I don't have a strong opinion. There's a nice convenience to automatically handling intermediate nodes (`mkdir -p`, `rm -r` - but notice they require special flags!). It might be tedious to have to do them all, one-by-one—round-trips get expensive in content-addressed land. But on the other hand, does it indicate a possible error if you're encountering missing intermediates? Perhaps so.

Your call if you're able to form a firmer opinion.

RangerMauve: I've been chewing on it and I think the convenience outweighs the downsides. 😅

I've got auto-generation (-p) for ipfs/ipns/hyper in Agregore at the moment and not having it made a lot of things more annoying. Ultimately, I think developer UX is really important for this stuff and streamlining is usually worth it.

RangerMauve:
I think this is good to merge 👍

PR #201, opened 2022-04-05
chore: update Eleventy and dependencies for 11ty@1.0.0

rvagg:
Mainly to deal with dependabot security warnings, but this is a bit overdue anyway. Updates all dependencies, the main big one being eleventy's 1.0.0 release, which has some minor breaking changes but the diff here seems to have taken care of the big ones (`sh` syntax highligher missing, .gitignore in .site no longer being read).

I ran a build, and the only material differences are some additional `<a>` elements inside headers now for named linking. Everything else seems to come out the same.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 node_modules
+.site/_output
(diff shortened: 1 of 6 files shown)


rvagg:
https://bafybeia3utfixc7zst4cua2xrelgmixpr3pypxhgnec7tczy5bx5o5zrpy.on.fleek.co/ is the preview, looks good to me

willscott (approved):

Issue #160, opened 2021-12-10
DAG-JSON spec isn't the most recent version


vmx:
It looks like something went wrong when merging the specs repo into this repo.

The DAG-JSON specification at https://ipld.io/specs/codecs/dag-json/spec/ isn't the most recent of the spec. The specification at https://github.com/ipld/specs/blob/2a106a9c22a570f436f92ca8aa3de85eb675e116/block-layer/codecs/dag-json.md contains more information and also contains the change that the base64 encoded binary bytes do *not* contain the Multibase prefix.

rvagg:
Going to take care of this soon, will also roll up closing out:

* https://github.com/ipld/ipld/issues/159
* https://github.com/ipld/ipld/issues/148

rvagg:
https://github.com/ipld/ipld/pull/212

Issue #148, opened 2021-11-10
dag-json slash escaping


tysonzero:
It seems like making dag-json truly support the full IPLD Data Model would not be particularly onerous.

One possibility is mapping `"/"` map keys in the data-model to `"//"`, `"//"` to `"///"` and so on.

Whilst this is potentially breaking, the situations in which it will break are situations that are playing with fire as is, by having keys very similar to the disallowed `"/"`.

rvagg:
ooof, it looks like the latest edits to the spec didn't make it in migration to this repo, but we do cover in detail some of the reserved space (`"/"`) rules @ https://github.com/ipld/specs/blob/master/block-layer/codecs/dag-json.md#the-reserved-namespace

These rules narrow down the range of unsupported data a little so it's not just that `"/"` isn't a supported key. The maps that use that as a reserved key are single-key maps, so >1 key shouldn't get caught in the trap (although I can't recall if our main implementations do this properly or not!).

But yes, you're right, we could introduce escaping and have discussed doing this. I don't recall why we didn't but it might have just been the additional complexity for a codec that we don't expect to have as wide ("serious") use as dag-cbor. But that could be a bad reason to not pursue it!

PR #212, opened 2022-05-10
fix: update dag-json spec with lots of clarifications

rvagg:
This is mostly existing work I did in the old ipld/specs repo but it was not ported here properly.

Closes: https://github.com/ipld/ipld/issues/160
Closes: https://github.com/ipld/ipld/issues/148


vmx (approved):
Thanks for tackling this!

PR #227, opened 2022-06-28
spec: bencode codec spec

aschmahmann:
This is a proposal for an IPLD Codec for the Bencode format. The Bencode format is commonly used within the BitTorrent ecosystem.

There's likely room for discussion here as to when someone would want to use a Bencode IPLD codec rather than some custom codec for formats like BitTorrent-infohash-v1 or BitTorrent-infohash-v2 that would also expose IPLD Links.

However, the existence of this format may help us explore more of the IPLD Data Model and also has utility in terms of being able to work with BitTorrent data without necessarily making lots of new codecs.

### DIFF
--- a/specs/codecs/bencode/spec.md
+++ b/specs/codecs/bencode/spec.md
@@ -1,0 +1,56 @@
+---
+title: "Bencode Specification"
+navTitle: "Spec"
+---
+
+# Specification: Bencode
+
+**Status: Prescriptive - Experimental**
+
+* [Format](#format)
+* [Implementations](#implementations)
+    * [Go](#go)
+    * [Rust/WASM](#rustwasm)
+
+[Bencode] is an existing format most commonly associated with the [BitTorrent] protocol. While it's both possible and likely that candidates working with BitTorrent data may want codecs that more specifically target those formats so as to properly resolve [Links] rather than leaving them as Bytes, being able to work with the Bencode data using IPLD seems reasonable as well.
+
+## Format
+
+From the [BitTorrent] v1 spec Bencode is defined as:
+
+- Strings are length-prefixed base ten followed by a colon and the string. For example 4:spam corresponds to 'spam'.
+- Integers are represented by an 'i' followed by the number in base 10 followed by an 'e'. For example i3e corresponds to 3 and i-3e corresponds to -3. Integers have no size limitation. i-0e is invalid. All encodings with a leading zero, such as i03e, are invalid, other than i0e, which of course corresponds to 0.
+- Lists are encoded as an 'l' followed by their elements (also bencoded) followed by an 'e'. For example l4:spam4:eggse corresponds to ['spam', 'eggs'].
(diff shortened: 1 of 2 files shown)


specs/codecs/bencode/spec.md
```diff
@@ -0,0 +1,56 @@
...
+- Lists are encoded as an 'l' followed by their elements (also bencoded) followed by an 'e'. For example l4:spam4:eggse corresponds to ['spam', 'eggs'].
+- Dictionaries are encoded as a 'd' followed by a list of alternating keys and their corresponding values followed by an 'e'. For example, d3:cow3:moo4:spam4:eggse corresponds to {'cow': 'moo', 'spam': 'eggs'} and d4:spaml1:a1:bee corresponds to {'spam': ['a', 'b']}. Keys must be strings and appear in sorted order (sorted as raw strings, not alphanumerics).
+
+### Strings
+
+Bencode Strings are represented as IPLD Data Model Strings. This is because, despite not being UTF-8 it makes it very simple to put a Bencode String into a Bencode Dictionary given that IPLD Data Model maps require String keys. As a result these Strings must take advantage of the IPLD Data Model flexibility to support non-UTF-8 data as Data Model Strings.
```
aschmahmann, line 28: @vmx here's an example where strings are not necessarily UTF-8 because bencode strings aren't UTF-8 and BitTorrent will even do things like put the concatenated SHA1 hashes and put them into a string.

At the moment the BitTorrent ADL (in Go and the further along one in Rust) is based off of the Bencode codec rather than a specific BitTorrent codec. There are some tradeoffs here, but wanted to flag how non-utf8 keys are used here since you were interested.

Some other materials related to this you might want to see where I mention some of the tradeoffs here before the demo towards the end.
- https://www.youtube.com/watch?v=Hc-5WJxBkpI
- https://discuss.ipfs.tech/t/2022-07-13-data-and-ipfs-models/14635/5

vmx: Thanks for pinging. Yes that's the problem of no real support for bytes in map keys. Possible workarounds could be:

 - Storing dictionaries as a list of two-tuples instead of maps
   - Con: this makes pathing ugly as you would need to know the position of key
 - If a bencode string is as map identifier, encode it as UTF-8 compatible string
   - Con: again pathing would be more difficult, you would need to know how to encode those binary bytes. On the other hand I guess current pathing implementations support UTF-8 only anyway, so you'd need to encode/escape the path somehow. You could use that as your "native" encoding.

RangerMauve: Regarding ugly pathing, I've written some ideas about how one could escape arbitrary binary data in paths using URL escaping with IPLD URLs. https://github.com/ipld/ipld/blob/e6cfab631d2bd24bf158d3a85e126514c98de5ce/notebook/exploration-reports/2022.03-ipld-url-scheme.md#special-characters-and-escaping

It'd be nice if Node implementations could support modes where keys are loaded as byte buffers. 😅

Issue #176, opened 2022-01-29
broken .md links


tabcat:
There are many links inside markdown files linking to other markdown files that are broken on github.

They are all only broken by a little bit; the links are just missing the `.md` extension. Here is an example of a [broken link inside the repo](https://github.com/ipld/ipld/blob/master/specs/selectors/index.md#fixtures). Many of these exist.

Happy to make a pr and comment of a script that handled repairing the links.

tabcat:
Was wondering if there was a reason for this. The links work on the ipld site. Adding `.md` to the paths result with an ipfs error since there is no link named FAQ.md under [bafybeidpga2xwiiseib3tcsp33wodu3itunwj2su624yygdx6av3s4f5f4](https://explore.ipld.io/#/explore/bafybeidpga2xwiiseib3tcsp33wodu3itunwj2su624yygdx6av3s4f5f4). For them to work on github and the website both would need to be changed to `.md`, but there may have been a reason for this in the first place.

rvagg:
Yeah, this annoys me a lot, I _think_ it's because eleventy, the static site generator, needs them in this broken state, at least that's been my experience authoring these things.

But if you can find a way to harmonise them nicely so we get a proper static HTML site at ipld.io _and_ working GitHub links (and also working editor links in VS Code and friends) then that'd be awesome!

BigLep:
@tabcat : contribution is welcome.  For now, we're giving priority to website viewing rather than markdown viewing given we don't know of an easy way to resolve this :(

rvagg:
Technical solution is provided by @dboreham and merged @ #232, the remaining item here is to work through the existing markdown files and rewrite their links so they all work. #232 only did it for a small subset.

tabcat:
Would like to write a script for this sometime in the near future. Seems like there are two cases to make edits:

- add `index.md` to links to folders
- add `.md` to links to files
- may need to adjust path relativity?

PR #230, opened 2022-07-25
Eth doc fixes

dboreham:
Fixed some dangling references in the eth spec docs.


rvagg:
yeah .. so I know this stucks but this is intentional to make the website build, it's a quirk of e11ty that we'd dearly love to address so these docs aren't so hostile in markdown, see #176

But as you can see @ https://ipld.io/specs/codecs/dag-eth/, they're currently working as they should for the website

sorry, but if you could help with #176 then that'd be awesome

PR #232, opened 2022-07-30
Fix broken .md links when viewed as markdown

dboreham:
Proposed fix for : https://github.com/ipld/ipld/issues/176

I found this [11ty bug report](https://github.com/11ty/eleventy/issues/1204)  which seems to be about the same core issue. Using some of the solutions presented there as inspiration I coded up this PR which is designed to address the narrow problem of our markdown files not properly linking when accessed via GitHub web interface, or a local md-aware editor.

I have included fixups to the _md_ _files_ _themselves_ only in one area : `specs/codecs/dag-eth`. There are many more instances remaining to be changed. Figured it was best to leave them until it's clear this approach is viable.

I have manually tested the result, checking that the web site served by `yarn start` has connecting links, and the relevant pages viewed in GitHub now link correctly. `yarn test` also passes.

As a drive-by fix I also changed the `permalink` configuration for `markdown-it-anchor` because build log output warned the current code was not valid, per : https://github.com/valeriangalliat/markdown-it-anchor . I am not however at all sure if the permalink scheme I picked is what we want. The change I made cures the build warning but I don't understand enough about what we want from permalinks to know if it's the correct thing.

### DIFF
--- a/.site/.eleventy.js
+++ b/.site/.eleventy.js
@@ -8,6 +8,7 @@ const markdownItContainer = require('markdown-it-container')
 const markdownItFootnote = require('markdown-it-footnote')
 const markdownItMark = require('markdown-it-mark')
 const markdownItTOC = require('markdown-it-table-of-contents')
+const markdownItReplaceLink = require('markdown-it-replace-link')
 const nunjucks = require('nunjucks')
 const prismIpldsch = require('./prism-ipldsch')
 const loadLanguages = require('prismjs/components/')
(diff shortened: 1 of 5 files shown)


rvagg:
Yeah, that seems to work! preview link is in the GitHub "checks" if you expand it: https://bafybeigv5uqd3ip35rbgf2rfcv4fmh4gjzbnpqroinlefjdfpidyvvl4h4.on.fleek.co/ is the current one for your latest commit and it looks good I think. I think the regexes and replacements make sense, I wonder if we'll hit weird corner cases but we have a link checker in here that isn't reporting problems with the compiled version.

I guess the next step is to find all of the broken links in the markdown versions, but this is nice because it doesn't require us be exhaustive now. Well done!

rvagg (approved):

dboreham:
Thanks! Looks like I don't have privs to merge though.

rvagg:
yeah, that's fine, I was just waiting to see if there's additional feedback before anyone else before jumping to the merge

PR #233, opened 2022-08-02
dag-pb: further clarifications on link sorting

rvagg:
This is shaping up to be a likely explanation for inconsistencies in Filecoin deals, with traversals yielding differently ordered blocks. I'm writing up more detailed notes on that case for an issue on https://github.com/filecoin-project/boost but the important thing here is to note that there _is_ a difference, and that we've moved almost all of the ecosystem away from sort-on-decode, but that if you want stable traversals then you shouldn't mix and match the two.

### DIFF
--- a/specs/codecs/dag-pb/spec.md
+++ b/specs/codecs/dag-pb/spec.md
@@ -11,6 +11,16 @@ DAG-PB is an IPLD codec that uses [Protocol Buffers](https://developers.google.c
 
 DAG-PB does not support the full [IPLD Data Model](/docs/data-model/).
 
+* [Implementations](#implementations)
+* [Serial Format](#serial-format)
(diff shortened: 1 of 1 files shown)


willscott (approved):

RangerMauve (approved):

rvagg:
FYI I'm sitting on this one for a moment, it turns out to be much more subtle than saying "before version X" and "after version X" of go-merkledag. There's current behaviours there that still need to be flagged (and probably fixed) so I should capture that accurately.

rvagg:
I've updated with more detail and referred to an as yet unreleased v0.6.0 of go-merkledag based on https://github.com/ipfs/go-merkledag/pull/87.

PR #239, opened 2022-09-07
DAB-CBOR Spec spelling correction

Winterhuman:
Corrects "limitaitons" to "limitations" in the DAG-CBOR spec page


RangerMauve (approved):
👍 Good catch, ty

PR #254, opened 2022-11-23
Added initial prolly tree ADL Spec

RangerMauve:
This spec isn't 100% complete yet, but it's at a point where we have enough figured out that we can get some final feedback on the details from folks that have deep IPLD knowledge.

This is based on the first milestone of [this devgrant](https://github.com/ipfs/devgrants/pull/194).

We have an MVP implementation where we sketched some code up based on Dolt just to see if we could get it working with IPLD at all. It's here and it has some tess and methods of rendering trees to graphviz. 

Once we finish this spec we'll reimplement stuff based on the spec phrasing to make sure it's compliant (and expose as an ADL). https://github.com/kenlabs/ptree-bs/pull/1

On the TODO list is:
- [ ] Finalize the spec
  - [ ] Should we configure the hash functions used for chunking, if so how
  - [ ] How should we do hashing of keys for chunking, namely is the `value` necessary to be hashed and if so what sort of limitations do we have with "large" values.
  - [ ] How should we deal with (or should we even) nodes that end up > 2MiB
  - [ ] Final touchups
- [ ] fixtures
- [ ] DMT for schemas
- [ ] ADL implementation and tests
- [ ] Rust implementation

cc @taoshengshi

### DIFF
--- a/specs/advanced-data-layouts/prollytree/spec.md
+++ b/specs/advanced-data-layouts/prollytree/spec.md
@@ -1,0 +1,467 @@
+# IPLD Prolly Trees Specification
+
+## Introduction
+
+Many applications have been using IPLD to represent large datasets in use cases such as blockchains or decentralized databases.
+One side effect is that querying large amounts of data has become a more common need which often gets solved with centralized database indexes.
+IPLD Prolly Trees are an important step in this direction in that they provide an interface similar to a Database's B+ tree.
+Some of the immediate benefits of using Prolly Trees over regular B+ Trees is that Prolly Trees are self-balancing and can be determenistically merged with other trees which is important to reduce the amount of restructuring necessary for collaborativ index creation.
+In this document we will build off of prior art in [Probabilistic B-Trees](https://github.com/attic-labs/noms/blob/master/doc/intro.md#prolly-trees-probabilistic-b-trees) to define a speficication on how they can be represented in IPLD, how they can be constructed, how they can be queried, and how multiple trees can be merged together.
+
+## References
+
+[DoltDB Architecture](https://github.com/dolthub/docs/tree/gitbook-publish/content/architecture)
+The best deep dive into how the Dolt storage engine works is a series of blog posts by Aaron Son.
+[How Dolt Stores Table Data](https://www.dolthub.com/blog/2020-04-01-how-dolt-stores-table-data/)
+
+[The Dolt Commit Graph and Structural Sharing]()
+https://www.dolthub.com/blog/2020-05-13-dolt-commit-graph-and-structural-sharing/)
+
+[Efficient Diff on Prolly Trees](https://www.dolthub.com/blog/2020-06-16-efficient-diff-on-prolly-trees/)
+
+[Cell-level Three-way Merge in Dolt](https://www.dolthub.com/blog/2020-07-15-three-way-merge/)
+
+[Dolt Implementation Notes — Push And Pull On a Merkle DAG](https://www.dolthub.com/blog/2020-09-09-push-pull-on-a-merkle-dag/)
+
+[mikeal/IPSQL: InterPlanetary SQL](https://github.com/mikeal/IPSQL)
+[mikeal/prolly-trees: Hash consistent search trees.](https://github.com/mikeal/prolly-trees)
+[mikeal/matrika: Next Generation Decentralized Database](https://github.com/mikeal/matrika)
+[mikeal/ipfs-sqlite: SQL on IPFS](https://github.com/mikeal/ipfs-sqlite)
+
+[Merkle Search Trees: Efficient State-Based CRDTs in Open Networks (Scientific Artilcle)](https://hal.inria.fr/hal-02303490)
+  - [simulation](https://gitlab.inria.fr/aauvolat/mst_exp/)
+
+## Summary/Overview
+
+Prolly trees leverage content addresibility to create ordered search indexes similar to the common B+ tree structure used for databases, but with the added ability to determenistically merge trees together.
+
+At the highest level, Prolly Trees act as a key value store whith the ability to iterate over key ranges which are alphanumerically sorted. This property of sorted iteration is important for creating database indexes for a variety of use cases like full text search, sorting of large datasets, and arbitrary search queries.
+
+### Search Tree
+
+The basic structure is that of an ordered search tree: The contained keys are organised such that they can be found (inserted, updated, ...) efficiently by value.
+
+*TODO Insert sketch of search tree*
+
+To efficiently find keys, the tree is traversed top-to-bottom and the non-leaf nodes help navigating/comparing the values efficiently. An intermediate node contains several ordered key-address pairs, which link to further nodes (intermediate or leaf) on the next lower level.
+
+Levels go from `0` representing Leaf Nodes, and go up for each level in the tree. The root of the tree will have the highest level in the tree and can give an estimate of it's overall size.
+
+Leaf nodes contain the actual Key-Value pairs for the tree which can be iterated over as part of the overall tree iteration.
+
+
+### Chunking
+
+Chunking is the strategy of determining chunk boundaries: Given a list of key-value pairs, it 'decides' which are still inside node A and which already go to the next node B on the same level.
+It depends on the hash/address/node of the items and a 'chunking factor' for tuning the shape of the tree.
+This chunking factor determines the average size of nodes/chunks that a node on a higher level contains.
+This average size in turn controls the shape of the tree (broad/narrow).
+The shape in turn defines the performance of operations on the tree.
+
+### Note on Multiple and Single Values and Sets
+
+The described tree can represent a data structure with multiple, single or no values per key. However, given that IPLD Maps (which a Prolly Tree loosely maps to) only allow one value for a key, implementations should merge duplicate keys into a single value. This is also important to have consistent ordering of key value pairs.
+
+### Re-Calculate
+
+In order to modify the tree (inserting, updating, or removing one or multiple values) a part of the tree needs to be re-calculated. This section describes an internal procedure that is not exposed to the caller.
+The tree is (partially) re-calculated bottom-up.
+
+Starting at the inserted/modified leaf or node referencing a now deleted/removed leaf, successively walk up the tree.If a removed item was the only item in its node: Remove the node from its parent and Check if the hash/address/CID of a new leaf/node splits the parent chunk (split the parent node if yes).
+    Then, check if a removed leaf/node was a 'splitting node' and the nodes need to be merged (only the last node in a chunk can be a splitting node). If splitting criterion holds for the last item of the node and there is the current node is succeeded by another node on the same level: merge the currend node with the succeeding node. Continue walking up the path. If at root and it is being split: Create a new root that links to the freshly split nodes. Return the new root CID.
+
+![](https://i.imgur.com/xE0id0V.png)
+![](https://i.imgur.com/aYqzXZI.png)
+![](https://i.imgur.com/nuvInkH.png)
+![](https://i.imgur.com/zACS8gS.png)
+
+Pay attention to the fact that the boundary algorithm is not related with the node cid, i.e node cid is got after the boundary is generated and the node is saved in blockstore. Maybe in the future we can combine the boundary algorithm and cid. 
+
+### Put/Remove/Update
+
+First search if the tree already contains the value.
+If the key is not contained, create a new leaf for the value, insert the key-value (key-leaf) pair in the correct node and re-calculate the tree.
+If the key is already stored in the tree and the value is the same, the tree is left unchanged.
+If the key exists and the values differ: Either leave the tree or insert the new value and update the tree.
+If removing the key, remove the `key` and the `value` from the list. After making any modifications to the tree, run the algoritm (Re-Calculate)[#Re-Calculate].
+
+## Structure
+
+IPLD Schema(https://github.com/kenlabs/ptree-bs/blob/main/pkg/prolly/tree/schema.ipldsch)
+
+```
+type ProllyNode struct {
+	config &ChunkConfig
+    level Int
+	keys [Bytes]
+	links nullable [&ProllyNode]
+	values nullable [Any]
+} representation tuple
+
+type WeibullThresholdConfig struct {
+	K Float
+	L Float
+} representation tuple
+
+type RollingHashConfig struct {
+	rollingHashWindow Int
+} representation tuple
+
+type PrefixThresholdConfig struct {
+  chunkingFactor Int
+}
+
+type ChunkStrategy enum {
+  | PrefixThreshold
+  | WeibullThreshold
+  | RollingHash
+} representation string
+
+type ChunkConfig struct {
+	chunkStrategy ChunkStrategy
+	minChunkSize Int
+	maxChunkSize Int
+	prefix nullable PrefixChunkConfig
+	weilbull nullable WeibullThresholdConfig
+	rollingHash nullable RollingHashConfig
+}
+```
+
+### `ProllyNode.keys`
+
+Raw keys(keys/values input from users) for leaf node. Key-value pairs are sorted by byte value with the "larger" keys being at the end. Values are comared at the first byte, and going down to the end. This means that keys that are just a prefix come before keys that are prefix + 1 byte.
+
+### `ProllyNode.values`
+
+raw values for leaf nodes. For branch nodes, it's null. Values can point to arbitrary IPLD nodes and it is up to applications to generate and process them.
+
+### `ProllyNode.links`
+
+null for leaf nodes. For branch nodes, it's the CID of the child node. The index in the array corresponds to the index in the `keys` array for the corresponding key. Keys are used for searching, and links are used for traversing down towards the leaves.
+
+### `ProllyNode.level`
+
+0 for leaf nodes, and add 1 for parent levels (and incrementing as more parents are added)
+
+### `ProllyNode.config`
+
+Link to the info about the chunking strategy for how the prolly tree is built. Trees must always be mutated with the given strategy. Storing the config in prolly nodes means that any parent inside a tree can be passed around as a new root, and it makes it easy to compare if two trees are compatible for being "merged" together or used for comparisons.
+
+### `ChunkStrategy`
+
+The enum for the types of chunking strategies that are part of the spec. This is set to be a string so that the number of strategies can grow over time.
+
+### `ChunkConfig`
+
+Chunk Config for prolly tree, it includes some global setting, the splitter method you choose and specific configs about the splitter
+
+### `ChunkConfig.minChunkSize`
+
+The minimum size a chunk should be before considering the boundry function.
+
+TODO: Should this be in number of keys or bytes of keys+values or block size? Maybe we can get away with number of keys?
+
+### `ChunkConfig.maxChunkSize`
+
+The maximum size a chunk could be before it needs to be split regardless of the chunk boundries
+
+### `ChunkConfig.chunkStrategy`
+
+The string representing the type of strategy to use. Either `RollingHash` or  `KeySplitter`.
+	
+### `PrefixThresholdConfig`
+
+Config for the `PrefixThreshold` chunking strategy.
+
+This is the original strategy that was described in the Merkle Search Tree paper and has a `chunkingFactor` which represents the number of bits in the key which need to be `0` in order to set another boundry.
+
+Lower values result in wider nodes.
+
+### `WeibullThresholdConfig`
+
+Config for the `WeiBullThreshold` chunking strategy.
+
+Makes chunk boundary decisions on the hash of the key of a []byte pair and tries to create chunks that have an average number of []byte pairs, rather than an average number of Bytes. However, because the target number of []byte pairs is computed directly from the chunk size and count, the practical difference in the distribution of chunk sizes is minimal. It uses a dynamic threshold modeled on a weibull distribution (https://en.wikipedia.org/wiki/Weibull_distribution). As the size of the current trunk increases, it becomes easier to pass the threshold, reducing the likelihood of forming very large or very small chunks.
+
+The `K` value represents the Shape Parameter, and `L` represents `λ` - the scaling factor.
+
+### `RollingHashConfig`
+
+ rollingHashSplitter is a nodeSplitter that makes chunk boundary decisions using a rolling value hasher that processes Item pairs in a byte-wise fashion.
+
+rollingHashSplitter uses a dynamic hash pattern designed to constrain the chunk Size distribution by reducing the likelihood of forming very large or very small chunks. As the Size of the current chunk grows, rollingHashSplitter changes the target pattern to make it easier to match. The result is a chunk Size distribution that is closer to a binomial distribution, rather than geometric.
+
+## Algorithm in detail
+
+Left to the implementation is how to generate and load data from IPLD.
+Whenever a Link is resolved to a ProllyNode or ChunkConfig, this is done via the IPLD LinkSystem which it an implementation detail outside of the scope of this document.
+
+When serializing data into a CID+Block, one should use the codec and multihash that's either used by root CID of the tree you are modifying, or specificed explicitly during tree creation. Blocks should also be saved to the IPLD LinkSystem and how this is done is also outside the scope of this document. We will assume that there is a `getNode(cid)` API which loads IPLD Nodes from a CID, and a `saveNode(node) => {cid, bytes}` which will save a node and get back a CID and the byte contents used for generating the CID. Note that `saveNode` should use the same encoding and hash algorithm as the root of the tree.
+
+### IsLeaf(ProllyNode) : Boolean
+
+1. get the `level` of the `ProllyNode`
+2. if the `level` is `0` return `true`
+3. else return `false`
+
+### CursorIsValid(Cursor) : Boolean
+
+1. get the `length` of `cursor.node.keys`
+2. if `length` is `0`, return `false`
+3. if `cursor.index` is less than 0, return `false`
+4. if `cursor.index` is greater than or equal to `legnth`, return `false`
+
+### KeyIndex(ProllyNode, item) : Integer
+
+Use a [binary search](https://en.wikipedia.org/wiki/Binary_search_algorithm) or equivalent to find the index in the `keys` array which is "closest" to but not "larger than" the `item`. Return the index.
+
+### CursorAtItem(ProllyNode, item): Cursor
+
+Get the closest cursor to a byte prefix
+
+1. Define a `Cursor` struct with a `index Integer`, `node ProllyNode`, and`parent Cursor`
+2. Set `cursor.node` to the `ProllyNode`
+3. Set `cursor.index` to `KeyIndex(cursor.node, item)`
+6. Start a loop
+	1. if `IsLeaf(cursor.node)` is `true`, break the loop
+	2. get the `link` from `CursorGetLink(cursor)`
+	3. resolve the ProllyNode at the `link` to `newNode`
+	4. Set `parent` to `cursor`
+	5. set `cursor` to a new Cursor struct
+	6. set `cursor.parent` to `parent`
+	7. set `cursor.node` to `newNode`
+	8. set `cursor.index` to `KeyIndex(cursor.node, item)`
+7. return the `cursor`
+
+### AdvanceCursor(Cursor) : Cursor
+
+1. Get the `length` of the `cursor.node.keys`
+2. if `cursor.index` is less than `length - 1`, increment `cursor.index` and return the `cursor`
+3. If `cursor.parent` is `null`
+	1. set `cursor.index` to `length`
+	2. return the `cursor`
+3. Invoke `AdvanceCursor(cursor.parent)`
+4. If `CursorIsValid(cursor.parent)` is `false`
+	5. set `cursor.index` to `length`
+	6. return the `cursor`
+6. Get the `link` from `CursorGetLink(cursor.parent)`
+7. Checgreater or k that `link` is not `null`, throw an error if it is
+8. Get the `node` ProllyNode from `link`
+9. Set `cursor.node` to `node`
+10. Set `cursor.index` to `0`
+
+### CursorGetKey(Cursor) : key?
+
+1. If `CursorIsValid(Cursor)` is `false`, return `null` (or throw an error)
+2. Get the `key` from `cursor.node.keys` at `index`
+3. return the `key`
+
+### CursorGetValue(Cursor) : value?
+
+1. If `IsLeaf(cursor.node)` is `false`, return `null` (or throw an error)
+1. If `CursorIsValid(Cursor)` is `false`, return `null` (or throw an error)
+2. Get the `value` from `cursor.node.values` at `cursor.index`
+3. return the `value`
+
+### CursorGetLink(Cursor) : &ProllyNode
+
+1. If `IsLeaf(cursor.node)` is `true`, return `null` (or throw an error)
+1. If `CursorIsValid(Cursor)` is `false`, return `null` (or throw an error)
+2. Get the `link` from `cursor.node.links` at `cursor.index`
+3. return the `link`
+
+### Get(ProllyNode, key): value
+
+1. Get a `cursor` from `CursorAtItem(ProllyNode, key)`
+2. If `CursorIsValid(cursor)` is `false`, return an error (key not found)
+3. Get `currentKey` from `CursorGetKey(Cursor)`
+4. If `currentKey` is not bytewise equal to `key`, return an error (key not found)
+5. Get `value` from `CursorGetValue(Cursor)`
+6. Return `value`
+
+### weibullCDF(x, K Int, L Int) : int
+
+Note: CDF (Cumulative Distribution Function) of the ][Weibull probability distribution](https://en.wikipedia.org/wiki/Weibull_distribution)
+ - return `-exp(-pow(x/L),K)`
+
+### WeibullThreshold(ProllyNode, key, value) : Boolean
+
+TODO: cleanup
+
+WeibullThreshold returns true if we should split at |hash| for a given record inserted into a chunk of Size |Size|, where the record's Size is |thisSize|. |Size| is the Size of the chunk after the record is inserted, so includes |thisSize| in it.
+
+WeibullThreshold attempts to form chunks whose sizes match the weibull distribution.
+
+The logic is as follows: given that we haven't split on any of the records up to |Size - thisSize|,  the probability that we should split on this record is (CDF(end) - CDF(start)) / (1 - CDF(start)), or, the precentage of the remaining portion of the CDF that this record actually covers. We split is |hash|, treated as a uniform random number between [0,1), is less than this percentage.
+
+0. Get the `{bytes, cid}` from `saveNode(value)`
+1. `hash = hash(key + saveNode(value).cid)` *TODO: hash needs to be well-defined*
+2. `itemSize = len(key) + len(value)`
+3. Set `prevItemsSize` to the sum of the length of all the node's previous items (excluding the current one)
+4. `start = weibullCDF(prevItemsSize, K, L)`
+5. `end = weibullCDF(prevItemsSize + itemSize, K, L)`
+6. `p = float(hash)/maxUInt32`
+7. `d = 1 - start`
+8. If `d <= 0`: return `true` (this should realistically never occour?)
+9. `target = (end - start)/d`
+10. return `p < target`
+
+### HashByteArray([]byte, offset) : Boolean
+
+*TODO: Description*
+*TODO: current implementation applies the min/maxChunkSize to offset. Do we want this?*
+- For each `byte`:
+    1. Increment `offset` by 1
+    2. Feed `byte` to rolling hash *TODO Details for hashing, salting*
+    3. Set `hash` to hashSum of rolling hash
+    4. `pattern = (1<<(15 - (offset >> 10))) - 1`
+    5. If `hash&pattern == pattern` return `true`
+- Return `false`
+
+### RollingHash(ProllyNode, index) : Boolean
+
+*TODO: Rolling hash window*
+1. Set `offset` to the number of bytes of the elements of previous keys and values
+2. If `HashByteArray(key) == true`: return `true`
+3. Add number of bytes of key to `offset`
+4. If `HashByteArray() == true`: return `true`
+5. return `false`
+
+### ShouldCreateBoundry(ProllyNode, key) : Boolean
+
+If `true`, a new ProllyNode should be created for all subsequent keys.
+
+1. Get the `ChunkConfig` by resolving the link in `node.ChunkConfig`
+1. Get the `ChunkStrategy` from the `ChunkConfig`
+2. Get the `length` from the `node.keys`
+3. If `length` is less than `ChunkConfig.MinChunkSize` return `true`
+3. If `length` is equal to `ChunkConfig.MaxChunkSize` return `true`
+4. If the `ChunkStrategy` is `KeySplitter`
+	- return WeibullThreshold(key)
+5. If the `ChunkStrategy` is `RollingHash`
+	- return RollingHash(ProllyNode, index)
+6. Return an error (unsupported chunk strategy)
+
+### RebalanceTree(Cursor) : ProllyNode
+
+The `cursor` should be pointing to the node in a tree that rebalancing should start at. The root of the `cursor` will be upated to match the changes.
+
+The returned `ProllyNode` is for the new root of the tree.
+
+TODO: Account for `cursor.parent` being null
+TODO: Advance a cursor instead of iterating through `cursor.node.keys`
+TODO: Account for `cursor.node.keys` being empty (remove from parent)
+
+- Loop
+  - If `ShouldCreateBoundry(cursor.node.chunkconfig, cursor.node, CursorGetKey(cursor))` is `false`
+	- Call `AdvanceCursor(cursor)`
+	- continue to next loop cycle
+  - Create a new `ProllyNode` `node` by cloning the current `cursor.node`
+  - Remove all keys, values, links from `cursor.node` after `cursor.index`
+  - Remove all keys, values, links from `node` before and including `cursor.index`
+  - Generate a new `CID` for `cursor.node`
+  - Set the `cursor.parent.links` at `cursor.parent.index` to the new `CID`
+  - Generate a new `CID` for the new `node`
+  - Shift all items in `cursor.parent.keys` up by `1`
+  - Shift all items in `cursor.parent.links` up by `1`
+  - Increment `cursor.parent.index` by `1`
+  - Set the `cursor.parent.links` at `cursor.parent.index` to the new `CID`
+  - Set the `cursor.parent.keys` at `cursor.parent.index` to node.keys[0]
+  - Set cursor.node to `node`
+  - Set cursor.index to `0`
+- Return `RebalanceTree(cursor.parent)`
+
+### Put(ProllyNode, key, value) : ProllyNode
+
+- Get `cursor` from `CursorAtItem(ProllyNode, key)`
+- If `CursorIsValid(cursor)` is `true`
+	- check if the key is at `CursorGetKey`
+	- If it is
+		- Set the value in the `cursor.node.values` at `cursor.index` to `value`
+	- If it isn't
+		- add the `key` to `cursor.node.keys` after the `cursor.index`, shifting subsequent items down
+		- add the `values` to `cursor.node.values` after the `cursor.index`, shifting subsequent items down
+- TODO: If false, how is this reached? It means that there were no keys even "close" to `key`?
+- return `RebalanceTree(cursor)`
+
+### Delete(ProllyNode, key, value) : ProllyNode
+
+- Get `cursor` from `CursorAtItem(ProllyNode, key)`
+- If `CursorIsValid(cursor)` is `false`
+	- Return the `ProllyNode`
+- check if the key is at `CursorGetKey(cursor)`
+- If it is
+	- Remove the key in `cursor.node.keys` at `cursor.index`
+	- Remove the value in `cursor.node.values` at `cursor.index`
+- return `RebalanceTree(cursor)`
+
+### Search(ProllyNode, prefix) : Iterator<key, value>
+
+- Get `cursor` from `CursorAtItem(ProllyNode, key)`
+- Create an Iterator (language dependent)
+- On each pull from the iterator
+	- If `CursorIsValid(cursor)` is `false`
+		- Close the iterator and return
+	- Get the `key` from `CursorGetKey(cursor)`
+	- If `key` does not start with `prefix`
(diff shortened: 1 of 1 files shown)


RangerMauve:
@SionoiS By Entree do you mean Entry (I think English and French have different spelling)? I'm down to add wording to that extent. Should make some of the wording more terse.

Mind elaborating on what you mean by "Doesn't matter if you hash stuff or not either. Multiple Entrees get then chunked and the tree built."

> Config should be optional

I'm still not sure about this since any intermediate node could be useful as a root. In particular if it contains a subset of the keyspace that's relevant for a particular use case.

matheus23:
Hi :wave: 

Following this, as I'm super interested in this as a possible solution for directory sharding in WNFS: https://github.com/wnfs-wg/spec/issues/8

> > Config should be optional
>
> I'm still not sure about this since any intermediate node could be useful as a root. In particular if it contains a subset of the keyspace that's relevant for a particular use case.

My thinking on this would be, there should be 'special' ("blessed") root nodes that contain the config.
It's unlikely that a specific intermediate node boundaries match *exactly* the boundaries someone else is looking for.

What I think is more likely is that there will be two prolly tree roots that share a subset of nodes. Perhaps because they were derived from the same original prolly tree or perhaps because one is a subset view of the other or something similar.

In these cases you end up with both of these trees starting in these blessed root nodes which can contain the config.

SionoiS:
> @SionoiS By Entree do you mean Entry (I think English and French have different spelling)? I'm down to add wording to that extent. Should make some of the wording more terse.

Yes your right, entry/entries as in dictionary entries or ledger entry, would be the right word.

> Mind elaborating on what you mean by "Doesn't matter if you hash stuff or not either. Multiple Entrees get then chunked and the tree built."

I just meant that keys can be anything that can be ordered.

Entries can be; key-value pairs, just keys if used as a set instead of a map or links to other blocks of the tree with some kind of data used for ordering.

> I'm still not sure about this since any intermediate node could be useful as a root. In particular if it contains a subset of the keyspace that's relevant for a particular use case.

Yes, but only the builder the the actual tree will know which subset is relevant, not us.
IMO the spec should say all non-root node CAN have a link to the config IF the user benefit from it, otherwise omit to reduce overhead.

> My thinking on this would be, there should be 'special' ("blessed") root nodes that contain the config.
> It's unlikely that a specific intermediate node boundaries match exactly the boundaries someone else is looking for.
> 
> What I think is more likely is that there will be two prolly tree roots that share a subset of nodes. Perhaps because they were derived from the same original prolly tree or perhaps because one is a subset view of the other or something similar.
> 
> In these cases you end up with both of these trees starting in these blessed root nodes which can contain the config.

Also, nothing prevent us from adding config to any new blocks as the tree is updated.

che-ch3:
> IMO the spec should say all non-root node CAN have a link to the config IF the user benefit from it, otherwise omit to reduce overhead.

Please keep in mind that the overhead is on _link_ (to a config) per node. I would like to argue that this is not at all that much.
Further I would argue that it simplifies data structures: No discerning between root/non-root node, no consideration of optional data within nodes. I personally highly appreciate simplicity in non-essential parts.
My gut feeling would be that operations on sub-trees happen all the time (as part of larger operations). Whether there's a need to keep the config at each node - I don't know yet.

More importantly, what will happen is that root nodes will become non-root nodes in newer trees (ex. when items are added). If we were to follow the strict distinction between root/non-root, we would have to change the prior root node into a non-root node. Which would mean that links that point to the old root node are invalidated. If we don't change/invalidate the old root, we need to make a 'copy' of the old root to insert it into the new tree, which would be against the idea of Merkle-ization. If we don't distinguish between root and non-root, we simply create the new root node, link from it to the old and everything stays valid.

SionoiS:
> More importantly, what will happen is that root nodes will become non-root nodes in newer trees (ex. when items are added). If we were to follow the strict distinction between root/non-root, we would have to change the prior root node into a non-root node. Which would mean that links that point to the old root node are invalidated. If we don't change/invalidate the old root, we need to make a 'copy' of the old root to insert it into the new tree, which would be against the idea of Merkle-ization. If we don't distinguish between root and non-root, we simply create the new root node, link from it to the old and everything stays valid.

That's a good point. I'm convinced! Link at every node makes more sense.

matheus23:
>  can folks answer the following questions?
>   * Does including the chunk config CID in every node instead of having official "root nodes" add enough overhead to matter?

> > More importantly, what will happen is that root nodes will become non-root nodes in newer trees (ex. when items are added). If we were to follow the strict distinction between root/non-root, we would have to change the prior root node into a non-root node. Which would mean that links that point to the old root node are invalidated. If we don't change/invalidate the old root, we need to make a 'copy' of the old root to insert it into the new tree, which would be against the idea of Merkle-ization. If we don't distinguish between root and non-root, we simply create the new root node, link from it to the old and everything stays valid.
> 
> That's a good point. I'm convinced! Link at every node makes more sense.

I'm not quite convinced. In newer trees, the old root will actually always be split into multiple nodes, so that the new root links to multiple children. I think the only case when the root node becomes a perfect child is if you're appending the "perfect" key value pairs to e.g. the end of the sequence, but I think that's fairly unlikely?

In any case - I don't think it's really worth arguing about. I wouldn't care much if there's a 40 byte overhead in each node. I'd be more concerned about the crosslinking perhaps, but even that tools should be able to handle.

>   * Should we configure the hash functions used for chunking, if so how

I'd say, today: Add some signaling. E.g. some integer identifying the chunking algorithm in a table that you make up specifically for this spec. The table may only have a single entry today, but implementations will know how to error out properly if they encounter a value they don't expect.
If in the future we come up with some specialized, perhaps parameterized way of defining a chunker, then so be it, we'll add another algorithm identifier and that may inform implementations to look for the parameters somewhere (in a way that keeps old implementations error out expectedly).

>   * How should we do hashing of keys for chunking, namely is the `value` necessary to be hashed and if so what sort of limitations do we have with "large" values.

Hash the keys, not the values, right? If you'd hash the keys then an in-place update of a value could change the chunking.
I'd expect most 'downstream users' of this prolly tree will instantiate values with
- CIDs
- Unit (empty struct/record) so it's a map
- Some other constant-sized thing

>   * How should we deal with (or should we even) nodes that end up > 2MiB

A developer should configure an entry limit per nodes. Similar to the 1000 entry limit for UnixFS directories, computed from the key-values given a max allowed directory name size.

The developer should choose a limit that allows the block to stay within their chosen block size, e.g. 262kB or 1MB.

That parameter informs the probability distribution function such that it cuts off at a certain point. Such a 'cutoff point' is inefficient, as you'll end up splitting and merging your node all the time, but that should be super rare given a well-chosen probability distribution function.

>   * Should we finalize all the chunking strategies at once, or limit to one with room for growth

The price we pay for adding an integer that signals the chunking strategy, even if it's just one is really low, IMO. So yeah, I'd say limit to one, but add that signaling so there's room to grow eventually.

>   * Should we merge the `values` and `links` into a single field and say that if level != 0 it must contain a CID to a ProllyNode.

:man_shrugging: Seems reasonable to me? Is there any value in linking to values at non-0 level?

>   * Should we suggest default encodings + hash functions + codecs for the CIDs/Blocks?

Perhaps? I wouldn't mind a lot if it's missing from the spec but then is mentioned in benchmarks/there are defaults in implementations.

RangerMauve:
Had a call with Mikeal today to talk about this. Some notes from what we talked about:

- It would be useful to have a general "ordered tree" spec where a reader can assume that the tree is sorted, and can then traverse it however it was created
- Prolly trees could be a superset of this base "tree" pair
- Having the config outside of tree nodes would be useful for cases where two "tree" specs create the same intermediate nodes or leaf nodes. In this case it would be good to have a "root" type which links to the config and the root prolly node.
- Hash functions should be in the config using the multihash code or similar, this will be used by writers when creating the prolly tree
- It would be useful to include the config in a custom multicodec (not sure how much effort this would be in practice)
- Values in leaves could be stored in `key, value` tuples for perf benefits
- We might want to consider supporting prolly trees that are not just Map analogues, but also something like append only logs where entries are appended at the end and there are no keys (personally I think it'd be useful to scope things to maps to start and look at this as a separate primitive, just due to priorities I have around the map use case)
- For block storage going over 1MiB, the spec doesn't need to have anything specific to guard against it, but we can mention about upper limits for number of entires and suggest configuration options for chunking that would enforce that. There might be larger block sizes in the future in new systems
- Parent or child node could just be a boolean, the level might not be necessary? (actually might be needed for some chunking strategies)
- Echoing that it'd be useful to start with one chunking strategy which uses the byte suffix of a hash (suffix because multihash usually uses first few bytes for codec reasons)
- Should leave space for future chunking strategies like weibull + rolling
- Hashing entries for chunking should be both the key and the value

I'll think more about this and some of the other comments and propose some spec revisions to what we have so far. Our current spec is based on what we learned from Dolt, but I think there's room for changes to fit with IPLD more cleanly.

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,467 @@
...
+
+## Algorithm in detail
+
+Left to the implementation is how to generate and load data from IPLD.
+Whenever a Link is resolved to a ProllyNode or ChunkConfig, this is done via the IPLD LinkSystem which it an implementation detail outside of the scope of this document.
+
+When serializing data into a CID+Block, one should use the codec and multihash that's either used by root CID of the tree you are modifying, or specificed explicitly during tree creation. Blocks should also be saved to the IPLD LinkSystem and how this is done is also outside the scope of this document. We will assume that there is a `getNode(cid)` API which loads IPLD Nodes from a CID, and a `saveNode(node) => {cid, bytes}` which will save a node and get back a CID and the byte contents used for generating the CID. Note that `saveNode` should use the same encoding and hash algorithm as the root of the tree.
+
+### IsLeaf(ProllyNode) : Boolean
```
sssion, line 201: the functions(include below) may be defined as structure functions? such as func(n *ProllyNode) IsLeaf() bool

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,467 @@
...
+1. get the `length` of `cursor.node.keys`
+2. if `length` is `0`, return `false`
+3. if `cursor.index` is less than 0, return `false`
+4. if `cursor.index` is greater than or equal to `legnth`, return `false`
+
+### KeyIndex(ProllyNode, item) : Integer
+
+Use a [binary search](https://en.wikipedia.org/wiki/Binary_search_algorithm) or equivalent to find the index in the `keys` array which is "closest" to but not "larger than" the `item`. Return the index.
+
+### CursorAtItem(ProllyNode, item): Cursor
```
sssion, line 218: nodeStore should be part of parameters for link resolving, maybe compareFunc(how to compare two []byte key) is also meaningful

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,467 @@
...
+### RollingHash(ProllyNode, index) : Boolean
+
+*TODO: Rolling hash window*
+1. Set `offset` to the number of bytes of the elements of previous keys and values
+2. If `HashByteArray(key) == true`: return `true`
+3. Add number of bytes of key to `offset`
+4. If `HashByteArray() == true`: return `true`
+5. return `false`
+
+### ShouldCreateBoundry(ProllyNode, key) : Boolean
```
sssion, line 330: I think we need a structure `Splitter`(ChunkSplitter) with state(such as offset, hash in RollingHash). We put k1v1, k2v2... into Splitter one by one and change the state then return whether boundary generates, if true we need reset the Splitter state and save the generated node.
We need a function to receive the input k/v one by one like append(k, v []byte). It must input in order, can't skip, repeat or draw back.
If only this function, it miss important information is where the computing starts. i.e, which range is the boundary generated for? Because I only see a pair of k/v is used in computing

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,467 @@
...
+  - Shift all items in `cursor.parent.keys` up by `1`
+  - Shift all items in `cursor.parent.links` up by `1`
+  - Increment `cursor.parent.index` by `1`
+  - Set the `cursor.parent.links` at `cursor.parent.index` to the new `CID`
+  - Set the `cursor.parent.keys` at `cursor.parent.index` to node.keys[0]
+  - Set cursor.node to `node`
+  - Set cursor.index to `0`
+- Return `RebalanceTree(cursor.parent)`
+
+### Put(ProllyNode, key, value) : ProllyNode
```
sssion, line 374: structure method,  and I think it's better to have another struct like ProllyTree?

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,467 @@
...
+	- If `CursorIsValid(cursor)` is `false`
+		- Close the iterator and return
+	- Get the `key` from `CursorGetKey(cursor)`
+	- If `key` does not start with `prefix`
+		- Close the iterator and return
+	- Get the `value` from `CursorGetValue(cursor)`
+	- Yield the `key` and `value` from the iterator
+	- AdvanceCursor(ProllyNode)
+
+### Diff(ProllyNode base, ProllyNode new) : Iterator< Diff >
```
sssion, line 412: Should the function be public api or just used in the merge function?

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,467 @@
...
+  chunkingFactor Int
+}
+
+type ChunkStrategy enum {
+  | PrefixThreshold
+  | WeibullThreshold
+  | RollingHash
+} representation string
+
+type ChunkConfig struct {
```
sssion, line 120: Would it be better to use union to replace the multi configs?
type ChunkConfig struct{
    chunkStrategy ChunkStrategy
    minChunkSize Int
    maxChunkSize Int
    strategy      strategy
} representation tuple

type strategy union {
    | WeibullThresholdConfig    "weibull"
    | RollingHashConfig         "rollinghash"
    | PrefixThresholdConfig     "prefix"
} representation keyed

type ChunkStrategy enum {
  | PrefixThreshold
  | WeibullThreshold
  | RollingHash
} representation string

specs/advanced-data-layouts/prollytree/spec.md
RangerMauve: I was thinking we could purposefully ignore any Object Oriented bits to the code so that it'd make as few assumptions as possible about the programming environment. Structure functions are very golang-y so it might be more specific than necessary for the spec.

Implementers can use these function signatures as vague guides they can use to write their code but are free to implement whatever function signatures they need. The important part is the shape of the data and making sure they consistently create the same trees from the same input.

specs/advanced-data-layouts/prollytree/spec.md
RangerMauve: If you check out the beginning of the "Algorithm in detail" section I mention how the APIs for saving/loading nodes is intentionally left abstract so that implementations can use whatever they want.

specs/advanced-data-layouts/prollytree/spec.md
RangerMauve: I don't think we need to perscribe whether something is public or not in the spec. But IMO it should be a public API in implementations since it's generally useful to have.

specs/advanced-data-layouts/prollytree/spec.md
RangerMauve: Maybe we can have different code paths for tracking state based on the chunking strategy? Or have another way to represent what state the chunking strategy is holding?

RangerMauve:
So, I've been talking to folks more and ruminating and have come up with some thoughts for the questions that were initially made:

> Does including the chunk config CID in every node instead of having official "root nodes" add enough overhead to matter?

One thing that really stuck out to me here was Mikeal's ideal to have the tree structure itself be more general purpose than just prolly trees. With that in mind, I think we should remove the Prolly part of branch and leaf nodes, and have a general purpose "tree" with keys and values. Then we can have a prolly root that links to the config and the tree root.

e.g.

```
type ProllyTree struct {
  config &ProllyTreeConfig
  root &TreeNode
}

type TreeNode struct {
  keys [Bytes]
  links [&TreeNode]
  values [Any]
  level Int
}
```

> Should we configure the hash functions used for chunking, if so how

We should include the hash function to be used in the chunking strategy config. In particular, using the byte from the [multihash table](https://github.com/multiformats/multihash) to represent it. One thing to keep in mind is that some chunking strategies require the ability to supply salt (like the Weibull or RollingHash). This means not all functions in the table will be viable for every strategy. This is probably something that will need to error out on write.

We should also include the codec and other CID stuff in the config since sniffing that from the root CID is kind of cumbersome and hacky. Having all the config options you need to encode / chunk a tree in a single object can make it easier to pass around during tree construction.

We should also take @sssion 's suggestion of using a keyed union representation of the configs.

IMO, it'd be safe to suggest using cbor to start unless folks have a better idea. We (folks working on the devgrant) were also talking about making a custom encoding with multicodec so that we can squeeze every bit of of performance out, but we're going to put that off for future work and stick to what's fastest to get working now to start.

> How should we do hashing of keys for chunking, namely is the value necessary to be hashed and if so what sort of limitations do we have with "large" values.

In order to support structures where the same key can appear multiple times, we should use both the key and the value in the hashing. It's also useful to introduce further randomness to the chunking to make it more difficult to create specially crafted keys that can overload the structure.

So, the hashing should look like `hash(key + encode(value))`. This can be optimized somewhat with using digests where you can feed chunks to them, so we can avoid concatenating the bytes.

I don't think we can prevent large values, but there should be wording in there about fixed sized value structures being the best option, and if you need large data, you should use a CID that points to the actual tree. This will also give us room for more fine grained optimization on writes and on storage down the line.

Regarding what @matheus23 said "Hash the keys, not the values, right? If you'd hash the keys then an in-place update of a value could change the chunking.". I'm actually not sure about why we shouldn't do this other than increasing randomness. @mikeal mentioned it was important but I honestly don't remember why other than avoiding some attacks and multi-key scenarios.

> How should we deal with (or should we even) nodes that end up > 2MiB

We should have some wording in the spec around "Hey, try to keep your blocks within this size, and here's some parameters that can help with that".

2 MiB is a limitation of the IPFS bitswap protocol, but it's not necessarily a limitation of all IPLD encoded blocks. With that in mind we should figure out some parameters to suggest folks to use as part of the spec.

One thing that we should consider is whether the `minChunkSize Int` and `maxChunkSize Int` are realistic for using with byte sizes and how to even calculate that without constantly re-encoding the block. My gut feeling is we'd be safe to start with a "maxKeys" and "minKeys" and punt the specific byte stuff down to chunking strategies that have more specific needs for it.

> Should we finalize all the chunking strategies at once, or limit to one with room for growth

Talking to folks, it sounds like having weibull and rolling hash down the line is useful, but that it'd also be useful to start with something simple like the chunking factor based on the hash of the key-value pair to start.

With that in mind, I'd propose to remove the existing weibull and rolling hash code, and leave room for it in subsequent prs when we add it to the list of functions. Ideally we'll start with the hash suffix and do some tests on it, then add weibull and rolling hash as we learn more. In particular, the weibull one seems useful for controlling the tree size in a more fine-grained manner that's more resistant to carefully crafted key-value pairs making the chunk too big.

One thing we might want to bikeshed is whether we should use a byte pointing to a table or a string. Personally I like strings since they're explicit and make it harder to have accidental collisions as we add new strategies. 

> Should we merge the values and links into a single field and say that if level != 0 it must contain a CID to a ProllyNode.

I'm still not sure about this one. The tradeoffs are that having an extra field will lead to a couple extra bytes, but will lead to slightly cleaner code and structs where it's obvious what part is values and what part can be expected to be prolly trees. It was also mentioned that we could get rid of the `level:0` heuristic for leaf nodes and inspect whether the value being pointed at is a TreeNode or not, but IMO the level doesn't add much overhead vs how complicated the code for that would end up being.

So it could either look like one of the following. IMO the first one is easier to make clean code interfaces for since you have an obvious mapping

```
type TreeNode {
  level Int
  keys [Bytes]
  links nullable [&TreeNode]
  values nullable [Any]
}
```

or

```
type TreeNode {
  level Int
  keys [Bytes]
  # If it's not a leaf it's a TreeNode
  values [Any]
}
```

Maybe we can also use a keyed union representation? I'm not sure if that's possible along with a tuple representation (which is useful for saving on bytes for the property names + makes it easier to use whatever property names you want).

I was also thinking that we could skip the `keys` and `values` arrays entirely and have an `entries` array with `[key, value]` tuples, but @sssion did some investigation and that would add a one or two byte overhead per entry compared to having two lists, and as you approach having 1000's of entries it adds up. Thus I don't think the entries approach is worth the overhead. As well, it might speed up searching through just the keys, but I have no data to back that up. This was done with the CBOR representaiton, but I wouldn't be surprised if similar stuff applied accross all codecs.

> Should we suggest default encodings + hash functions + codecs for the CIDs/Blocks?

We should at least provide one viable config for folks to test out and test their implementations against.

vmx:
@RangerMauve, this is a reply to your comment. To keep it shorter, I'll only comment on the things I care about and don't agree with.

> One thing that we should consider is whether the `minChunkSize Int` and `maxChunkSize Int` are realistic for using with byte sizes and how to even calculate that without constantly re-encoding the block

At least for the `maxChunkSize` I'd be in favour of having it in bytes. I'd rather make the implementation a bit more complicated, than ending up with consumers that and up with non-bitswappable blocks. For implementations, I could imagine that they have some threshold and they would split rather early before the `maxChunkSize` is reached and then error in case their heuristic was wrong, and then end up with re-encoding.

> > Should we merge the values and links into a single field and say that if level != 0 it must contain a CID to a ProllyNode.
> 
> ```
> type TreeNode {
>   level Int
>   keys [Bytes]
>   # If it's not a leaf it's a TreeNode
>   values [Any]
> }
> ```

I'm strongly in favour of this one. Having only `values` and not `links` matches how nodes in trees are usually defined (from my experience). You'd then have some indicator on whether it's an inner or leaf node. Using the level for that sounds useful to me. It may be used for things like statistics or diffing/merging.

vmx (commented):
Good work, that's really a good start.

I only had a quick look at the algorithms, I'd hope someone else would have a deeper look.

In addition to my inline comments I've a few general ones.

For the function definitions, I think it would be great to have a short description on what they actually do. Kind of the things you would put into your API docs.

I'm a bit confused about the images. They don't really seam to build upon each other, yet there are posted in sequence without further text. It's not really clear what each of them tries to explain (only partially).

At the beginning there are some references, but I think it would be great to have a section that explains the general idea which role the hashing plays for the chunking and why it leads to a deterministic tree.

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,467 @@
...
+
+### Re-Calculate
+
+In order to modify the tree (inserting, updating, or removing one or multiple values) a part of the tree needs to be re-calculated. This section describes an internal procedure that is not exposed to the caller.
+The tree is (partially) re-calculated bottom-up.
+
+Starting at the inserted/modified leaf or node referencing a now deleted/removed leaf, successively walk up the tree.If a removed item was the only item in its node: Remove the node from its parent and Check if the hash/address/CID of a new leaf/node splits the parent chunk (split the parent node if yes).
+    Then, check if a removed leaf/node was a 'splitting node' and the nodes need to be merged (only the last node in a chunk can be a splitting node). If splitting criterion holds for the last item of the node and there is the current node is succeeded by another node on the same level: merge the currend node with the succeeding node. Continue walking up the path. If at root and it is being split: Create a new root that links to the freshly split nodes. Return the new root CID.
+
+![](https://i.imgur.com/xE0id0V.png)
```
sssion, line 73: Should we remove these pictures?

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,467 @@
...
+ 3. Return `merged`.
+
+#### Note on Merge Conflicts
+
+When merging trees it is possible that the trees have different values for the same key. In this case, as the algorithm is ignorant of keys and values, it can not be known how to solve the conflict.
+The following approaches are possible for the implementation:
+ - Ask the caller for a handler function that resolves the conflict
+ - Throw an error, indicating the conflicting keys and values
+ - Ignore both
+ - Assume the `right` tree contains the newer values and simply use the key-value pair of the right tree.
```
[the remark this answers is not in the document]
RangerMauve: Could you elaborate on what we should have here instead?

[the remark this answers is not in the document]
vmx: The former two ways (error or a function) and not the latter two.

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,467 @@
...
+
+### KeyIndex(ProllyNode, item) : Integer
+
+Use a [binary search](https://en.wikipedia.org/wiki/Binary_search_algorithm) or equivalent to find the index in the `keys` array which is "closest" to but not "larger than" the `item`. Return the index.
+
+### CursorAtItem(ProllyNode, item): Cursor
+
+Get the closest cursor to a byte prefix
+
+1. Define a `Cursor` struct with a `index Integer`, `node ProllyNode`, and`parent Cursor`
```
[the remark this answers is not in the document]
RangerMauve: My only hesitation there is that I don't think cursors are something integral to the spec working and are more of an implementation detail. 😅 Honestly I was considering looking for a way to remove them from the spec text entirely.

Should I maybe note what a Cursor is and mention explicitly that it's a suggestion and not strictly necessary?

specs/advanced-data-layouts/prollytree/spec.md
RangerMauve: Yeah sure if they aren't adding much to the spec anyway. 👍

specs/advanced-data-layouts/prollytree/spec.md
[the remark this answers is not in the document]
vmx: Better wording from my side would've been: don't do implicit conversion, make it always in an explicit way the user has to deal with it, being it an error or a custom function.

specs/advanced-data-layouts/prollytree/spec.md
RangerMauve: Yeah, if we use a keyed union we can replace the `ChunkStrategy` enum. This would mean having unique key names per each strategy which might be annoying if we need reuse. 🤔 Let's try it out though.

specs/advanced-data-layouts/prollytree/spec.md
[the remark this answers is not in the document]
RangerMauve: Added more of a definition in the intro to the section.

RangerMauve:
K, pending a final review to find bugs and formatting errors, I think this is good to go as the final algorithm.

I got rid of the extra chunking strategies so we can stuck to the one in the peer to peer ordered search indexes blog post to start and changed how the chunking strategy is represented to account for vmx's comments.

Also did some refactoring of the spec code to account for edge cases and be more clear about min/max chunk config and how the boundry is calculated.

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,496 @@
+# IPLD Prolly Trees Specification
+
+## Introduction
+
+Many applications have been using IPLD to represent large datasets in use cases such as blockchains or decentralized databases.
+One side effect is that querying large amounts of data has become a more common need which often gets solved with centralized database indexes.
+IPLD Prolly Trees are an important step in this direction in that they provide an interface similar to a Database's B+ tree.
+Some of the immediate benefits of using Prolly Trees over regular B+ Trees is that Prolly Trees are self-balancing and can be determenistically merged with other trees which is important to reduce the amount of restructuring necessary for collaborativ index creation.
```
SionoiS, line 8: determenistically -> deterministically

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,496 @@
...
+In this document we will build off of prior art in [Probabilistic B-Trees](https://github.com/attic-labs/noms/blob/master/doc/intro.md#prolly-trees-probabilistic-b-trees) to define a speficication on how they can be represented in IPLD, how they can be constructed, how they can be queried, and how multiple trees can be merged together.
```
SionoiS, line 9: speficication -> specification

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,496 @@
...
+[mikeal/prolly-trees: Hash consistent search trees.](https://github.com/mikeal/prolly-trees)
+[mikeal/matrika: Next Generation Decentralized Database](https://github.com/mikeal/matrika)
+[mikeal/ipfs-sqlite: SQL on IPFS](https://github.com/mikeal/ipfs-sqlite)
+
+[Merkle Search Trees: Efficient State-Based CRDTs in Open Networks (Scientific Artilcle)](https://hal.inria.fr/hal-02303490)
+  - [simulation](https://gitlab.inria.fr/aauvolat/mst_exp/)
+
+## Summary/Overview
+
+Prolly trees leverage content addresibility to create ordered search indexes similar to the common B+ tree structure used for databases, but with the added ability to determenistically merge trees together.
```
SionoiS, line 36: addresibility -> addressability

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,496 @@
...
+
+### `ProllyTree`
+
+A prolly tree is identified by a node which links to the `root` `TreeNode` for the actual key-value pairs,
+and a `config` link to the ProllyTreeConfig which has infomration about the chunking and encoding information in order to quickly compare if two trees are using the same configuration and may be merged together as well as making it easier to write.
+
+### `TreeNode`
+
+This is the "Tree" part of Prolly Trees and is made to be general purpose.
+We can potentially expec to use the TreeNode structure in subsequent specs with different types of trees.
```
SionoiS, line 122: expec -> expect

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,496 @@
...
+It works by hashing a key+value pair, reading the last 4 bytes as a 32 bit unsigned integer, and checking how many bits are 0's relative to the chunking factor.
+
+The `chunkingFactor` must be less than the maximum value of a 32 bit unsigned integer.
+It is used to calculate a "chunking threshold" using the forumala `Math.floor(4294967295 / chunkingFactor)`.
+It is reccommended to use powers of two to make it easier to relate to how many bits should be `0` in the chunking threshold.
+
+The larger the chunking factor, the less likely it is that a given keypair will result in a chunking boundry, and thus will lead to TreeNodes with more entries within them.
+
+It also contains a `hashFunction` field which points at a hash function in the [multicodec table](https://github.com/multiformats/multicodec/blob/master/table.csv) to use for hashing.
+You should ensure that you are using a cryptographic hash function in order to make it more difficult to create collissions and to ensure you have an even distribution of values.
```
SionoiS, line 189: collissions -> collisions

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,496 @@
...
+[mikeal/ipfs-sqlite: SQL on IPFS](https://github.com/mikeal/ipfs-sqlite)
+
+[Merkle Search Trees: Efficient State-Based CRDTs in Open Networks (Scientific Artilcle)](https://hal.inria.fr/hal-02303490)
+  - [simulation](https://gitlab.inria.fr/aauvolat/mst_exp/)
+
+## Summary/Overview
+
+Prolly trees leverage content addresibility to create ordered search indexes similar to the common B+ tree structure used for databases, but with the added ability to determenistically merge trees together.
+
+At the highest level, Prolly Trees act as a key value store whith the ability to iterate over key ranges which are lexicographically sorted. This property of sorted iteration is important for creating database indexes for a variety of use cases like full text search, sorting of large datasets, and arbitrary search queries.
```
SionoiS, line 38: whith -> with

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,496 @@
...
+
+### Rebalance
+
+In order to modify the tree (inserting, updating, or removing one or multiple values) a part of the tree needs to be rebalanced. This section describes an internal procedure that is not exposed to the caller.
+The tree is (partially) rebalance bottom-up.
+
+Starting at the inserted/modified leaf or node referencing a now deleted/removed leaf, successively walk up the tree. If a removed item was the only item in its node: Remove the node from its parent and Check if the hash/address/CID of a new leaf/node splits the parent chunk (split the parent node if yes).
+Then, check if a removed leaf/node was a 'splitting node' and the nodes need to be merged (only the last node in a chunk can be a splitting node). If splitting criterion holds for the last item of the node and there is the current node is succeeded by another node on the same level: merge the currend node with the succeeding node. Continue walking up the path. If at root and it is being split: Create a new root that links to the freshly split nodes. Return the new root CID.
```
SionoiS, line 67: currend -> current

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -35,369 +35,401 @@ https://www.dolthub.com/blog/2020-05-13-dolt-commit-graph-and-structural-sharing
...
+### CursorAtItem(TreeNode, prefix): Cursor
 
-1. get the `level` of the `ProllyNode`
+Get a cursor pointing to the closest item in the tree to a given prefix.
+This is useful for performing searches for keys.
+
+1. Define a `Cursor` struct with a `index Integer`, `node TreeNode`, and`parent Cursor`
+2. Set `cursor.node` to the `TreeNode`
+3. Set `cursor.index` to `KeyIndex(cursor.node, prefix)`
+6. Start a loop
```
vmx, line 213:
```suggestion
4. Start a loop
```

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -35,369 +35,401 @@ https://www.dolthub.com/blog/2020-05-13-dolt-commit-graph-and-structural-sharing
...
+	1. if `IsLeaf(cursor.node)` is `true`, break the loop
+	2. get the `link` from `CursorGetValue(cursor)`
+	3. resolve the TreeNode at the `link` to `newNode`
+	4. Set `parent` to `cursor`
+	5. set `cursor` to a new `Cursor` struct
+	6. set `cursor.parent` to `parent`
+	7. set `cursor.node` to `newNode`
+	8. set `cursor.index` to `KeyIndex(cursor.node, item)`
+7. return the `cursor`
```
vmx, line 222:
```suggestion
5. return the `cursor`
```

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -35,369 +35,401 @@ https://www.dolthub.com/blog/2020-05-13-dolt-commit-graph-and-structural-sharing
...
 
 ### CursorIsValid(Cursor) : Boolean
 
+Check if a given cursor is set to a valid position.
+It can sometimes be set to an invalid position if a search failed or if there are no more items to iterator over in a cursor.
+
 1. get the `length` of `cursor.node.keys`
 2. if `length` is `0`, return `false`
 3. if `cursor.index` is less than 0, return `false`
 4. if `cursor.index` is greater than or equal to `legnth`, return `false`
```
vmx, line 240:
```suggestion
4. if `cursor.index` is greater than or equal to `length`, return `false`
```

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -35,369 +35,401 @@ https://www.dolthub.com/blog/2020-05-13-dolt-commit-graph-and-structural-sharing
...
 
-### KeyIndex(ProllyNode, item) : Integer
+### CursorIsAtEnd(Cursor) : Boolean
 
-Use a [binary search](https://en.wikipedia.org/wiki/Binary_search_algorithm) or equivalent to find the index in the `keys` array which is "closest" to but not "larger than" the `item`. Return the index.
+Check if the given cursor is at the end of it's TreeNode.
```
vmx, line 244:
```suggestion
Check if the given cursor is at the end of its TreeNode.
```

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -35,369 +35,401 @@ https://www.dolthub.com/blog/2020-05-13-dolt-commit-graph-and-structural-sharing
...
 	- Remove the key in `cursor.node.keys` at `cursor.index`
 	- Remove the value in `cursor.node.values` at `cursor.index`
-- return `RebalanceTree(cursor)`
+- get a new `root` from  `RebalanceTree(cursor, config)`
+- Create a new `updatedTree` by duplicating `tree`
+- Set `updatedTree.root` to `save(root).cid`
+- Return `updatedTree`
 
-### Search(ProllyNode, prefix) : Iterator<key, value>
+### Search(TreeNode, prefix) : Iterator<key, value>
```
vmx, line 426: There is still the problem that you cannot iterate over arbitrary ranges. The problem is the step `Close the iterator and return`. It will automatically stop when the prefix doesn't match, so it's not fully up to the application.

I propose changing `prefix` to `start` and don't automatically stop somewhere, but just keep iterating to the end, unless the application stops the iteration.

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -35,369 +35,401 @@ https://www.dolthub.com/blog/2020-05-13-dolt-commit-graph-and-structural-sharing
...
 
-The `K` value represents the Shape Parameter, and `L` represents `λ` - the scaling factor.
+This is the strategy that was described in the [Peer to Peer Ordered Search Indexes](https://0fps.net/2020/12/19/peer-to-peer-ordered-search-indexes/) paper.
 
-### `RollingHashConfig`
+It works by hashing a key+value pair, reading the last 4 bytes as a 32 bit unsigned integer, and checking how many bits are 0's relative to the chunking factor.
 
- rollingHashSplitter is a nodeSplitter that makes chunk boundary decisions using a rolling value hasher that processes Item pairs in a byte-wise fashion.
+The `chunkingFactor` must be less than the maximum value of a 32 bit unsigned integer.
+It is used to calculate a "chunking threshold" using the forumala `Math.floor(4294967295 / chunkingFactor)`.
```
vmx, line 183:
```suggestion
It is used to calculate a "chunking threshold" using the formula `Math.floor(4294967295 / chunkingFactor)`.
```

specs/advanced-data-layouts/prollytree/spec.md
RangerMauve: Sounds good 👍

I've mostly been using prefix based search with stuff like leveldb, but I can see the point in focusing on start instead of prefix.

RangerMauve:
K! I don't think we should merge this yet, but I think the spec is ready for a final review by @mikeal for the devgrant while we work on the initial ADL implementation. Ty to everyone that helped figure out the little details!

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,501 @@
...
+
+## Structure
+
+```
+type ProllyTree struct {
+  config &ProllyTreeConfig
+  root &TreeNode
+} representation tuple
+
+type TreeeNode struct {
```
che-ch3, line 87:
```suggestion
type TreeNode struct {
```
I guess the extra "e" was unintentional.

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,501 @@
...
+
+A prolly tree is identified by a node which links to the `root` `TreeNode` for the actual key-value pairs,
+and a `config` link to the ProllyTreeConfig which has infomration about the chunking and encoding information in order to quickly compare if two trees are using the same configuration and may be merged together as well as making it easier to write.
+
+### `TreeNode`
+
+This is the "Tree" part of Prolly Trees and is made to be general purpose.
+We can potentially expect to use the TreeNode structure in subsequent specs with different types of trees.
+
+### `TreeeNode.keys`
```
che-ch3, line 124:
```suggestion
### `TreeNode.keys`
```

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,501 @@
...
+If the TreeNode isn't a leaf, the function will traverse into it to find the leaf which contains the key.
+
+1. Get a `cursor` from `CursorAtItem(TreeNode, key)`
+2. If `CursorIsValid(cursor)` is `false`, return an error (key not found)
+3. Get `currentKey` from `CursorGetKey(Cursor)`
+4. If `currentKey` is not bytewise equal to `key`, return an error (key not found)
+5. Get `value` from `CursorGetValue(Cursor)`
+6. Return `value`
+
+### CursorAtChunkingBoundry(ProllyTreeConfig config, Cursor) : Boolean
```
che-ch3, line 307:
```suggestion
### CursorAtChunkingBoundary(ProllyTreeConfig config, Cursor) : Boolean
```

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,501 @@
...
+
+```
+type ProllyTree struct {
+  config &ProllyTreeConfig
+  root &TreeNode
+} representation tuple
+
+type TreeNode struct {
+  # Is leaf when level is 0
+  level Int
```
mikeal, line 89: I really don't think the level should go into the data.

It can't be trusted from the point of the view of the reader, and for the writer who is using this info to determine chunker settings they'll always know the depth because you've gotta write it from the branches up.

You do still need a way to differentiate branch from leaf nodes though, either as a type difference or a boolean in the data.

RangerMauve:
So, doing the differentiation via types isn't easy to do with how IPLD Schema works right now, so instead we'll change `level` to be an `isLeaf` boolean flag. 

With this in place the spec should be good to go for now and we'll leave it open as we get the Rust and Golang ADL implementations in place 👍

We might make changes to the spec as part of our learnings from those implementations, but it's likely the general structure won't change until we add new chunking strategies. 🎉

burdiyan:
I’m super excited about this spec. I’ve been looking into Prolly Trees for some time to see if they would work for what we do at [Mintter](https://github.com/mintterteam).

Have you looked at [Merkle Search Trees](https://hal.inria.fr/hal-02303490/document) which are [used by Bluesky](https://atproto.com/guides/data-repos#data-layout)? It seems like they are very similar to Prolly Trees, if not exactly the same (I haven’t read that paper deeply yet). Maybe it’s worth mentioning the differences/similarities between these in the spec?

RangerMauve:
> Have you looked at [Merkle Search Trees](https://hal.inria.fr/hal-02303490/document) which are [used by Bluesky](https://atproto.com/guides/data-repos#data-layout)?

Prolly trees are based on merkle search trees. The main difference is that Merkle Search Trees assign a "level" to the key-value pair based on the hash, and Prolly Trees assign chunk boundries based on the key-value pair hash. In general prolly trees are better for sequential reads.

Not sure about the approach bluesky is taking, but I'd love to collaborate with them on this. 👍

RangerMauve:
Golang ADL based on the spec is here: https://github.com/kenlabs/go-ipld-prolly-trees/pull/7/files#diff-7123ca137150a938ea6af380a465add25710f8f58e1b99c3a5187e505bf992b1

BigLep:
2023-01-24 maintainer conversation: we're going to let the implementation move forward and inform the spec.

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,508 @@
...
+It depends on the hash/address/node of the items and a 'chunking factor' for tuning the shape of the tree.
+This chunking factor determines the average size of nodes/chunks that a node on a higher level contains.
+This average size in turn controls the shape of the tree (broad/narrow).
+The shape in turn defines the performance of operations on the tree with the following tradeoffs:
+  - larger blocks with more cache invalidations
+  - or smaller blocks with less cache invalidations and more lookups
+
+### Note on Multiple and Single Values and Sets
+
+The described tree can represent a data structure with multiple, single or no values per key. However, given that IPLD Maps (which a Prolly Tree loosely maps to) only allow one value for a key, implementations should merge duplicate keys into a single value. This is also important to have consistent ordering of key value pairs.
```
garbados, line 59: Is there a recommended approach to merging duplicate keys? If I have a database index over a recipe database for "contains meat", my queries will return rows organized by a boolean key, thus being identical for some or even all result rows.

RangerMauve: I think this is very dependent on the application.

For some applications you can treat the value as a sort of CRDT and merge them, for some you might want a last write wins with some sort of clock, in some cases you might want to have both key value pairs beside each other and iterate over them both when doing a search.

For indexes I usually have the document's primary key in the index key so that they can be unique.

Here's where I did stuff in my hyperbee based DB which is also a key-value store: https://github.com/RangerMauve/hyperbeedeebee/blob/default/index.js#L782

aboodman:
> Prolly trees are based on merkle search trees. 

Little nit: prolly trees predate MSTs by several years.

RangerMauve:
@aboodman Snap, good to know. Mind sending some links for info on that so I can add them to the spec text?

aboodman:
They were first documented in 2016: https://github.com/attic-labs/noms/commit/8abe3a6fd90b5c7ec6957017a441bc6db30ebdcd#diff-dd95875d468bbbdd06a921cef4ad250818b906f448f642e868ade571f8a26c96R117

The implementation in that repo dates to probably the beginning of 2016 or late 2015.

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,508 @@
...
+
+This is the "Tree" part of Prolly Trees and is made to be general purpose.
+We can potentially expect to use the TreeNode structure in subsequent specs with different types of trees.
+
+### `TreeNode.keys`
+
+Raw keys(keys/values input from users) for leaf node. Key-value pairs are sorted by byte value with the "larger" keys being at the end. Values are compared at the first byte, and going down to the end. This means that keys that are just a prefix come before keys that are prefix + 1 byte.
+
+### `TreeNode.values`
+
```
SionoiS, line 130: Not sure about what is said here.

"Values corresponding to keys."
"Values can point to arbitrary IPLD nodes and it is up to applications to generate and process them."
"For non-leaf node these will be Links to other nodes."

Would make more sense IMO.

specs/advanced-data-layouts/prollytree/spec.md
```diff
@@ -0,0 +1,508 @@
...
+### `ProllyTreeConfig.hashFunction`
+
+This is the multicodec code for the hash function to use for generating CIDs.
+It is reccommended to use SHA2-256 for your hash function unless you know what you're doing.
+
+### `ProllyTreeConfig.hashLength`
+
+This is the multihash length parameter which should be used for generating CIDs.
+It can be set to `null` to use the default hash length from the hash function output.
```
SionoiS, line 174: What does null represent here? Zero?

AFAIK a tuple representation cannot have null/optional value.

maceip:
cc @tabcat

PR #264, opened 2023-01-06
fix: fix decode strictness links in dag-cbor spec

TripleDogDare:
Markdown links were backwards. Pretty straight-forward.

The @ipld/reviewers tag seems broken so I'm not sure if that still works and is the correct method for grabbing attention.


rvagg (approved):
thanks!

PR #284, opened 2023-05-06
Update dag-cbor spec.md - swap True and False (!)

DavidBuchanan314:
This swaps the values to match the definitions and ordering given in RFC-8949

https://datatracker.ietf.org/doc/html/rfc8949#fpnoconttbl2

I believe this was just a typo in the dag-cbor spec, and current implementations all do what RFC-8949 *actually* says (I think, I haven't checked exhaustively)


vmx (approved):
Good catch, thanks a lot!

rvagg (approved):
yeah, good catch, thanks!

PR #290, opened 2023-07-05
feat(trustless-pathing): add fixtures and go accessor pkg

rvagg:
Experimenting with parking pathing fixtures in this repo so it's not in lassie or repeated elsewhere. The main file in here is 20 Mb and currently 55 test cases in [Testmark](https://github.com/warpfork/go-testmark) format that use it in various ways to path through UnixFS data, but more to be added.

This comes out of https://github.com/filecoin-project/lassie/pull/306. There's also an associated tool that was originally used to generate these and can be used to print these descriptive paths through the data according to the queries, I'll get that online too but in a separate repo that can contain various non-test "trustless pathing" tooling to be consumed by lassie, frisbii and others.


hannahhoward:
LGTM

PR #294, opened 2023-08-16
feat(trustless-pathing): negative byte range fixtures

rvagg:
Ref: https://specs.ipfs.tech/http-gateways/trustless-gateway/#entity-bytes-request-query-parameter


hannahhoward (approved):
LGTMs

PR #296, opened 2023-08-24
chore: publish as @ipld/specs

rvagg:
/cc @guanzo

One problem with this is that there's no README, I need to work out if putting one in that directory is going to be a problem for site publishing. There's a README in pkg-js/trustless-pathing/ though that explains how to use that bit.

I haven't actually tried to use this, I'm doing a best-guess at what a package should look like. Care to help get it right @guanzo && @alanshaw?

### DIFF
--- a/specs/pkg-js/trustless-pathing/testcase.js
+++ b/specs/pkg-js/trustless-pathing/testcase.js
@@ -1,89 +1,82 @@
 import { CID } from 'multiformats/cid'
 
 class TestCase {
-    constructor(name, root, path, scope, duplicates, byteRange, expectedCids) {
-        this.name = name
-        this.root = root
-        this.path = path
-        this.scope = scope
-        this.duplicates = duplicates
-        this.byteRange = byteRange
-        this.expectedCids = expectedCids
-    }
+  constructor (name, root, path, scope, duplicates, byteRange, expectedCids) {
+    this.name = name
+    this.root = root
+    this.path = path
+    this.scope = scope
+    this.duplicates = duplicates
+    this.byteRange = byteRange
+    this.expectedCids = expectedCids
+  }
 
-    asQuery() {
-        let pp = this.path.split('/').filter(Boolean).join('/')
-        if (pp !== '') {
-            pp = '/' + pp
-        }
-        let br = ''
-        if (this.byteRange !== '') {
-            br = `&entity-bytes=${this.byteRange}`
-        }
-        let dup = ''
-        if (this.duplicates) {
-            dup = '&dups=y'
-        }
-        return `/ipfs/${this.root}${pp}?dag-scope=${this.scope}${br}${dup}`
+  asQuery () {
+    let pp = this.path.split('/').filter(Boolean).join('/')
+    if (pp !== '') {
+      pp = '/' + pp
     }
+    const br = this.byteRange ? `&entity-bytes=${this.byteRange}` : ''
+    const dup = this.duplicates ? '&dups=y' : ''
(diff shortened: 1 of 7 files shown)


alanshaw (approved):
🚢🚢🚢

rvagg:
added a basic readme and some ignores from site building

rvagg:
https://www.npmjs.com/package/@ipld/specs ok, well _something_ is up; the link in the readme doesn't work, so there's that.

rvagg:
Currently encountering problems decoding this unixfs data using ipfs-unixfs-exporter cause of in-built assumptions that we're dealing with hamt's with a width of 256: https://github.com/ipfs/js-ipfs-unixfs/blob/47218799b4c28b0a4b4dc5a71b9d981e5d803424/packages/ipfs-unixfs-exporter/src/resolvers/unixfs-v1/content/hamt-sharded-directory.ts#L27 - this data is synthetic and squishes up the hamts to get collisions and depth in a small number of blocks and assumes the implementation will work that out. ipfs-unixfs-exporter needs to read `fanout` from the unixfs data section and use that to work out how many characters to slice off the front of a link name, e.g. https://github.com/ipfs/go-unixfsnode/blob/7db26c0c9ffb19a4aba4299e0e2e4425d5598633/hamt/util.go#L87-L89

specs/pkg-js/trustless-pathing/testcase.js
```diff
@@ -0,0 +1,89 @@
...
+        this.expectedCids = expectedCids
+    }
+
+    asQuery() {
+        let pp = this.path.split('/').filter(Boolean).join('/')
+        if (pp !== '') {
+            pp = '/' + pp
+        }
+        let br = ''
+        if (this.byteRange !== '') {
```
alanshaw, line 20: Will be string or `null` after get from `URLSearchParams`.

```suggestion
        if (this.byteRange) {
```

specs/pkg-js/trustless-pathing/unixfs_20m_variety.js
```diff
@@ -0,0 +1,30 @@
...
+  return new URL(file + typ, import.meta.url)
+}
+
+export function unixfs20mVarietyCar() {
+  return pathToFixture('car')
+}
+
+export async function unixfs20mVarietyCases() {
+  const file = pathToFixture('md')
+  const contents = await readFile(file, 'utf-8')
```
alanshaw, line 17: Synchronous allows consumers to call and iterate test cases to define individual tests at the top level.

```suggestion
export function unixfs20mVarietyCases() {
  const file = pathToFixture('md')
  const contents = readFileSync(file, 'utf-8')
```

rvagg:
published a v1.0.2 with those changes just now @alanshaw

specs/pkg-js/trustless-pathing/testcase.js
```diff
@@ -0,0 +1,82 @@
...
+  let lines = exec.split('\n')
+  while (lines.length > 0 && lines[0].trim() === '') {
+    lines = lines.slice(1)
+  }
+  while (lines.length > 0 && lines[lines.length - 1].trim() === '') {
+    lines = lines.slice(0, lines.length - 1)
+  }
+  const specParts = spec.trim().split('?')
+  if (specParts.length !== 2) {
+    return new Error('invalid spec')
```
alanshaw, line 35: Did you mean to return these? Should be throw? Note: multiple places.

rvagg: go-brain probably, banged this out pretty quick

PR #305, opened 2023-09-20
Multiple text corrections

criadoperez:
Multiple text corrections on markdown files


willscott (approved):
Thanks for the copy-editing!

PR #306, opened 2023-09-20
Makefile and Glosary

criadoperez:
* In `make schema-schema` a space between "@" and "echo" can cause issues on older versions of `make` and it's best practice not to have it
* Removed duplicate `signalling` in Glossary
* Corrected a few typos


rvagg (approved):
neat! thanks

PR #308, opened 2023-09-25
chore: more detail & cleanup execution explanations

rvagg:
Does not impact usage of these tests as they just use the first column, the CID, which hasn't changed in this PR. This is cosmetic changes to match the tool which I wrote to produce these which I've since improved a little.


willscott (approved):

rvagg:
now it's published, https://github.com/ipld/go-fixtureplate is where these originally came from and the improvements I made prior to publishing are what's being reflected in this PR

PR #309, opened 2023-10-12
fix(trustless-pathing): 0:0 test for a sharded file

rvagg:


rvagg:
To clarify what's going on here: the existing `0:0` test doesn't cover the sharded file case, it asserts that the raw leaf file will be returned, which is obvious for an `entity` scope request, but in the sharded file case it should behave differently, we want the `File` block as well as its first `RawLeaf`.

Because this isn't being exercised, we've ended up with this bug: https://github.com/ipld/go-trustless-utils/pull/19, which looks like this, vs the fix below it:

Screenshot 2023-10-12 at 2 06 14 pm

PR #310, opened 2023-12-06
fix: DAG-CBOR keys need to be CBOR strings

vmx:
DAG-CBOR must be valid CBOR, hence the keys in maps need to be CBOR strings and not IPLD Data Model strings, which theoretically allow for arbitrary bytes.

### DIFF
--- a/specs/codecs/dag-cbor/spec.md
+++ b/specs/codecs/dag-cbor/spec.md
@@ -41,7 +41,7 @@ The inclusion of the Multibase prefix exists for historical reasons and the iden
 
 ## Map Keys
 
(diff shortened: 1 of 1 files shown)


rvagg:
CBOR maps allow all sorts of things for keys though, numbers, tagged values, even arrays or other maps. The point here is to restrict the valid keyspace down to what we recognise in the IPLD Data Model; which mainly comes down to major-3 entries, i.e. strings.

vmx:
Yes exactly, which is why it says "In DAG-CBOR, map keys must be strings.".

rvagg (approved):
ok then, I guess?

PR #323, opened 2024-03-25
docs: add note about exponent notation to DAG-JSON

vmx:
In JSON numbers can be represented with the exponent notation, hence represent two identical numbers the same way, e.g. `300` as `3e2`. Add a note that for content addressing you might want to restrict it to just decimals.

---

If someone comes up with better wording please do. I just thought mentioning it at least would be good.


rvagg (approved):

MITCHELLNEAL1:
> In JSON numbers can be represented with the exponent notation, hence represent two identical numbers the same way, e.g. `300` as `3e2`. Add a note that for content addressing you might want to restrict it to just decimals.
> 
> If someone comes up with better wording please do. I just thought mentioning it at least would be good.

PR #326, opened 2024-05-17
DagJose JSON payload support

oed:
Currently dag-jose only support payloads that are *base64url* encoded CIDs. This is a problem since most JOSE tools only support signing over JSON payloads. This PR extends the dag-jose codec to support arbitrary JSON payloads.

- [js-dag-jose impl. PR](https://github.com/ceramicnetwork/js-dag-jose/pull/32)
- [rust-dag-jose (WIP) PR](https://github.com/ceramicnetwork/rust-dag-jose/pull/40)

### DIFF
--- a/specs/codecs/dag-jose/fixtures/index.md
+++ b/specs/codecs/dag-jose/fixtures/index.md
@@ -22,7 +22,7 @@ Fixtures
 
 ### JWS
 
-This is the base64url-encoded CID for a DAG-JOSE object, when using SHA2-256 (multihash code 0x12):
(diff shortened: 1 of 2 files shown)


vmx (approved):
I'm not deep enough into DagJose, but it sounds like it's just describing what it's doing and it sounds reasonable. Though if someone who knows more about it (@rvagg maybe?) could have a look that would be great.

So I'd appreciate one more pair of eyes, I'll approve it though in case no one else finds the time time have a look.

rvagg (approved):

PR #356, opened 2025-09-03
dag-cbor: remove note about map keys sort order

vmx:
[It turns out] that the [canonical CBOR map key order in RFC 7049] and the [CBOR core deterministic encodings map key order of RFC 8949] are equal *as long* as the map keys are restricted to string only. This is the case in DAG-CBOR. Hence the note about the canonical order of RFC 7049 can be removed.

The requirement for string keys was changed to "must" in the Format section of this spec, as it is also a must in the "Map Keys" section.

[It turns out]: https://github.com/darobin/dasl.ing/issues/26#issuecomment-2967315011
[canonical CBOR map key order in RFC 7049]: https://datatracker.ietf.org/doc/html/rfc7049#section-3.9
[CBOR core deterministic encodings map key order of RFC 8949]: https://datatracker.ietf.org/doc/html/rfc8949#section-4.2.1

---

I've tried my best wording this in an understandable way, but if anyone has better ideas, please comment.

### DIFF
--- a/specs/codecs/dag-cbor/spec.md
+++ b/specs/codecs/dag-cbor/spec.md
@@ -28,7 +28,7 @@ DAG-CBOR uses the [Concise Binary Object Representation (CBOR)] data format, def
 The CBOR IPLD format is called DAG-CBOR to disambiguate it from regular CBOR. Most simple CBOR objects are valid DAG-CBOR. The primary differences are:
 
   * Tag `42` interpreted as CIDs, no other tags are supported.
-  * Maps may only be keyed by strings.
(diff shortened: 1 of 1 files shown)


rvagg (approved):
OK, cool, but these "must"s and "must reject"s aren't done in practice anywhere on the decode side, we're very lax and don't even have full strict modes (started a strict mode in Go at one point).

vmx:
> OK, cool, but these "must"s and "must reject"s aren't done in practice anywhere on the decode side

`serde_ipld_dag_cbor` only works with string keys, I would also assume on the decode side.

PR #380, opened 2026-07-22
dag-cbor: exclude -0.0

vmx:
A floating point value of `-0.0` is equal to `0.0` according to the IEEE 754 standard. So we have a number with different byte representatons. Hence always encode it as `0x0000000000000000`.


rvagg (approved):
fine by me; in JS we never produce -0.0 or -0 on encode but will sloppily deal with on decode, so the "should" still works for current state

but .. in go-ipld-prime we'll allow a -0.0 to flow through to the forbidden bytes; if you're going to block encode in Rust then maybe we should do the same in Go, would you mind opening an issue if/when you get to this?

PR #383, opened 2026-08-25
Allow data-first field order in dag-pb PBNode message

achingbrain:
Writing the `Links` first makes reading HAMT data more expensive since, if you are only trying to read a certain path, you have to read all of the `Links` before you can process any of them, as the config needed to calculate the hash prefix length is stored in the `Data` field at the end of the message.

Allow writing fields the other way round round so `Data` can be read before any `Links`.

Further discussion: https://github.com/ipfs/specs/issues/533
IPIP: https://github.com/ipfs/specs/pull/550

### DIFF
--- a/specs/codecs/dag-pb/spec.md
+++ b/specs/codecs/dag-pb/spec.md
@@ -51,11 +51,11 @@ message PBLink {
 }
 
 message PBNode {
-  // opaque user data
-  optional bytes Data = 1;
-
   // refs to other objects
   repeated PBLink Links = 2;
+
+  // opaque user data
+  optional bytes Data = 1;
 }
 ```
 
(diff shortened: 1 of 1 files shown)


rvagg:
hah, you're going to go through the whole IPIP-0499 CID profiles hoo-haa over this one too? a new line in the permutations table of how to represent things, a new 2026 profile?

lidel:
That is a good point, https://github.com/ipfs/specs/pull/550 should also define 2026 profile, without new profile we create breaking change to existing users. Documented in https://github.com/ipfs/specs/pull/550#discussion_r3853471792

Kubuxu:
I remember manually modifying the ordering in code generated by gogo-protobuf back in 2017 to preserve compatibility with this weirdness. :rofl:

achingbrain:
> hah, you're going to go through the whole IPIP-0499 CID profiles hoo-haa over this one too?

Thankfully the `unixfs-v1-2026` profile is just a one-setting change (so far, at least), and most of the hard work previously was aligning Kubo and Helia so hopefully it should be simpler this time around 😅 

Any obejctions to merging this and https://github.com/ipld/js-dag-pb/pull/111 ?

rvagg:
well, I don't love this because it's a breaking change, and my preference would be to make it opt-in rather than opt-out so you don't end up surprising people with new CIDs
you're also going to need to do the Rust and Go versions too and come up with a config strategy for them, they're not quite as easy as passing in an `options`

lidel:
Opt-in feels safer for everyone.

FWIW it is how it ended up on the UnixFS side: ipfs/specs#550 puts `data-first` behind a new `unixfs-v1-2026` profile (implicit default for existing users = links-first, as before), Go write support sits behind the same opt-in in ipfs/boxo#1212 and ipfs/kubo#11439, and [gateway-conformance v0.14.1](https://github.com/ipfs/gateway-conformance/releases/tag/v0.14.1) ships fixtures + tests for reading both orders. 

go-codec-dagpb itself is untouched and still writes links-first (the opt-in encoder lives in boxo for now, could be upstreamed as an encode option if you want it there: https://github.com/ipld/go-codec-dagpb/pull/103). 

So maybe this PR doesn't need to swap the canonical order at all? 

The swap is the breaking part. Keeping links-first as the implicit legact default encode order and documenting data-first as a valid opt-in variant (decoders must accept both anyway) would match what implementations ended up doing in proposed Go PRs.

achingbrain:
I've made it opt-in, though I think we should be brave in fixing the mistakes of the past, otherwise the only way to improve the ecosystem is to invent new things which leaves legacy code everywhere.

vmx:
@achingbrain could you please add a fixture to https://github.com/ipld/codec-fixtures/? That's a place where we test things across implementations. This way I can make sure that the Rust version decodes those blocks correctly.

achingbrain:
@vmx see https://github.com/ipld/codec-fixtures/pull/211 though it's not complete, further discussion on that PR

vmx (approved):

PR #384, opened 2026-08-25
fix dag-pb spec link

achingbrain:
Update link to point to the `.md` file, currently this just 404s.


rvagg (approved):
oh there was some bug a while back about index references, maybe this didn't get fixed in the sweep, or maybe we reverted something that undid it .. I hope there's not more of these

